Issue #628, opened 2021-03-09
Improve prometheus metrics


Pantani:
- [x] Create a metrics packages
- [x] Application Metrics (error/panic rates);
- [x] Database connections (Connectivity, Latency);
- [x] API requests (node API requests);
- [x] 3trd-party HTTP requests (prof server);
- [x] Host-Based Metrics (CPU, Memory, Disk space, Processes);

arnaubennassar:
@Pantani I see that all the items on the list are marked as done, can you close this issue?

PR #636, opened 2021-03-11
instrumenting the application

Pantani:
### What does this MR does?

Instrumenting the node with exp vars for Prometheus can collect and store.

- Create the metric package;
- Create a gin middleware to measure the API requests;
- Collect API error responses;
- Create a Prometheus collector to instrument the database connections;
- Create a method to measure the method execution time;
- Measure the proof server request;
- Add block reorg count metric;

### Why we need it?

For we have better node metrics for Prometheus to show inside the Grafana dashboard.

### How to test?

- run the node with the debug API on:
```
$ go run ./cli/node --mode coord --cfg cli/node/cfg.buidler.toml
``` 

- check the metrics `/debug/metrics`;

closes https://github.com/hermeznetwork/hermez-node/issues/628

#### Dashboard sample:
Screen Shot 2021-03-19 at 01 05 57


metric/request.go
```diff
@@ -0,0 +1,79 @@
...
+		reqCnt: reqCnt,
+		reqDur: reqDur,
+	}, nil
+}
+
+// PrometheusMiddleware creates the prometheus collector and
+// defines status handler function for the middleware
+func PrometheusMiddleware() gin.HandlerFunc {
+	p, err := NewPrometheus()
+	if err != nil {
```
ed255, line 56: if `err != nil` then `p == nil`.  If `p == nil` the middleware will panic when dereferencing `p`, which is inconvenient, because it happens later on once an API request is done, it happens in a goroutine, gin-gonic may recover the panic, etc.  I suggest this function either returns error, or panics, instead of just logging the error.

ed255 (changes requested):

ed255 (approved):

arnaubennassar (approved):

PR #668, opened 2021-03-23
Faster synchronization with less updates of Ethereum stats

OBrezhniev:
When doing initial synchronization we don't need to update Ethereum stats (fetching last Ethereum block mined and last batch forged) after each block (but not more that once in N second specified in config).

This change does this update only once in `cfg.Synchronizer.StatsUpdateFrequencyDivider` blocks synchronized. After reaching a threshold of `cfg.Synchronizer.StatsUpdateBlockNumDiffThreshold` blocks left to synchronize, node switches to regular checks on each block.

### DIFF
--- a/synchronizer/synchronizer.go
+++ b/synchronizer/synchronizer.go
@@ -68,11 +68,11 @@ var (
 // Stats of the synchronizer
 type Stats struct {
 	Eth struct {
-		RefreshPeriod time.Duration
-		Updated       time.Time
-		FirstBlockNum int64
-		LastBlock     common.Block
-		LastBatchNum  int64
+		UpdateBlockNumDiffThreshold uint16
+		UpdateFrequencyDivider      uint16
+		FirstBlockNum               int64
+		LastBlock                   common.Block
+		LastBatchNum                int64
 	}
 	Sync struct {
 		Updated   time.Time
@@ -107,9 +107,10 @@ type StatsHolder struct {
 }
 
 // NewStatsHolder creates a new StatsHolder
-func NewStatsHolder(firstBlockNum int64, refreshPeriod time.Duration) *StatsHolder {
+func NewStatsHolder(firstBlockNum int64, updateBlockNumDiffThreshold uint16, updateFrequencyDivider uint16) *StatsHolder {
 	stats := Stats{}
-	stats.Eth.RefreshPeriod = refreshPeriod
+	stats.Eth.UpdateBlockNumDiffThreshold = updateBlockNumDiffThreshold
+	stats.Eth.UpdateFrequencyDivider = updateFrequencyDivider
 	stats.Eth.FirstBlockNum = firstBlockNum
 	stats.Sync.LastForgeL1TxsNum = -1
 	return &StatsHolder{Stats: stats}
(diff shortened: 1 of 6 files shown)


synchronizer/synchronizer.go
```diff
@@ -23,6 +23,12 @@ const (
 	// errStrUnknownBlock is the string returned by geth when querying an
 	// unknown block
 	errStrUnknownBlock = "unknown block"
+	// updateEthBlockNumThreshold is a threshold of number of ethereum blocks left to synchronize, such that
+	// if we have more blocks to sync than the defined value we can aggressively skip calling UpdateEth
+	updateEthBlockNumThreshold = 100
+	// While having more blocks to sync than updateEthBlockNumThreshold, UpdateEth will be called once in a
```
Pantani, line 29:
```suggestion
	// While having more blocks to sync than updateEthBlockNumThreshold, UpdateEth will be called once in a
```
```suggestion
	// updateEthFrequencyDivider While having more blocks to sync than updateEthBlockNumThreshold, UpdateEth will be called once in a
```

synchronizer/synchronizer.go
```diff
@@ -23,6 +23,12 @@ const (
...
+	// defined number of blocks
+	updateEthFrequencyDivider = 100
```
Pantani, line 31: maybe it's better this value can be configurable in the config file by the owner

Pantani (changes requested):

synchronizer/synchronizer.go
ed255: I agree that these two parameters could be configurable

ed255:
Looks good! I think this approach is better than updating the eth stats only after a timeout, because it allows to skip calls when we haven't reached the current block, and update the stats at every block once we are in sync.

So I propose you delete the logic in `UpdateEth` that handles the timeout to decide if the update is made or not (https://github.com/hermeznetwork/hermez-node/blob/e68c64db759de9e8b1c88995ef9f39221d6ea048/synchronizer/synchronizer.go#L118), and also remove the configuration parameter `StatsRefreshPeriod`.

synchronizer/synchronizer.go
```diff
@@ -528,8 +534,11 @@ func (s *Synchronizer) Sync(ctx context.Context,
 	log.Debugf("ethBlock: num: %v, parent: %v, hash: %v",
 		ethBlock.Num, ethBlock.ParentHash.String(), ethBlock.Hash.String())
 
-	if err := s.stats.UpdateEth(s.ethClient); err != nil {
-		return nil, nil, tracerr.Wrap(err)
+	if nextBlockNum+updateEthBlockNumThreshold >= s.stats.Eth.LastBlock.Num ||
```
ed255, line 537: I suggest removing the time-based update approach in favor of this one. See https://github.com/hermeznetwork/hermez-node/pull/668#issuecomment-805780693

ed255 (changes requested):

synchronizer/synchronizer.go
OBrezhniev: Ok, I'm evaluating if this gives tangible performance gain after switching to fetching block headers.

Pantani (approved):

ed255 (approved):

Issue #663, opened 2021-03-22
Public API mock server is responding 404


arnaubennassar:
The swagger UI deployed at https://apidoc.hermez.network has an option to try the endpoints using a mock server.

However since we introduced versioning on the API, we're not using swagger versioning properly, which makes the documentation a bit poor and also breaks the paths of the mock server

**Edit:** note that updating the `api/swagger.yml` should also fix the server automatically (once the changes reach master)

arnaubennassar:
It looks like the swagger spec is correct but the mock server doesn't follow the spec: https://github.com/stoplightio/prism/issues/1591

I'll look for an alternative

arnaubennassar:
Closing in favor of #809

PR #669, opened 2021-03-23
Fix mock server

arnaubennassar:
Added `/v1` in swagger endpoints. This goes against the [OpenAPI spec](https://swagger.io/docs/specification/api-host-and-base-path/), however the mock server we're using doesn't support this (see this issue for more info #663).

IMO it's not bad to have the version added to the path in the description as this is actually part of the path and must be included, which is easy to miss if this i stetted only in the server URL. However not following a spec is a bad idea in general.

Anyway, as this is used publicly [here](https://apidoc.hermez.network/), we need a fix soon, and I've opted for this solution.
Note that my first attempt has been to try similar software, and for different reasons (problems with CORS, problems with the actual specification file, ...) non of them has worked.

After fixing the public mock server we should consider a more solid solution for this problem like adding a reverse proxy to manipulate the paths to be consistent with the OpenAPI spec, or just choose to ignore this.

EDIT: the original idea of this PR was to fix the mock server. Instead it switches to using the deployed API on testnet. After editing this comment the previously related issue won't get closed with the idea to fix the mock up server later on in order to give users the decision of using mocked data vs real (testnet) data

### DIFF
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -59,9 +59,9 @@ externalDocs:
   description: Find out more about Hermez network.
   url: 'https://hermez.io'
 servers:
-  - description: Hosted mock up
-    url: https://apimock.hermez.network/v1
-  - description: Localhost mock Up
+  - description: Testnet api, connected to Rinkeby deployment
+    url: https://api.testnet.hermez.io/v1
+  - description: Localhost usefull for testing/developing the api
     url: http://localhost:4010/v1
 tags:
   - name: Coordinator


api/swagger.yml
```diff
@@ -59,17 +59,21 @@ externalDocs:
...
+  - description: Mainnet (Ethereum) server, use it carefully, specially if attempting to send transactions. You could lose money!
+    url: https://api.hermez.io
 tags:
   - name: Coordinator
     description: Endpoints used by the nodes running in coordinator mode. They are used to interact with the network.
   - name: Explorer
     description: Endpoints used by the nodes running in explorer mode. They are used to get information of the netwrok.
 paths:
-  '/account-creation-authorization':
+  '/v1/account-creation-authorization':
```
arnaucube, line 76: Checking what it says at the spec that you referenced (https://swagger.io/docs/specification/api-host-and-base-path/), `v1/` should be part of the `server URL` and not of the `endpoint`:
```
    https://api.example.com/v1/users?role=admin&status=active
    \________________________/\____/ \______________________/
             server URL       endpoint    query parameters
                                path
```

api/swagger.yml
```diff
@@ -59,17 +59,21 @@ externalDocs:
...
-  - description: Localhost mock Up
-    url: http://localhost:4010/v1
+  - description: Hosted mock up, returns fake data usefull for development
+    url: https://apimock.hermez.network
+  - description: Localhost mock up, returns fake data usefull for development
+    url: http://localhost:4010
+  - description: Testnet (Rinkeby) server
+    url: https://api.testnet.hermez.io
+  - description: Mainnet (Ethereum) server, use it carefully, specially if attempting to send transactions. You could lose money!
+    url: https://api.hermez.io
```
arnaucube, line 69: I'm not sure if this information should go here. I understand this swagger config file as the file to run the mock api locally, but not as the directory of the current available deployed servers.
I think that `Hosted mockup`, `Testnet server` and `Mainnet server` should not go to this file, and should go to the documentation website.

arnaucube (changes requested):

api/swagger.yml
arnaubennassar: the idea is that the UI is actually a nice playground of the API, and I thought that it was a good idea to let people play with real data

api/swagger.yml
arnaubennassar: Ok, in this case I'd prefere to add a reverse proxy in the `docker-compose` to fix this, as I've already tried many different options and didn't liked any of them.

Also, another option is to completley remove the scripts needed to spin up the documentation and mock server from this repo. Now that the repo is open it should be straight forward

api/swagger.yml
Pantani: I'm not too fond of the idea of putting these config URLs into the swagger file. It's easy to lost track and forget to update. We can create a script to deploy it with a dynamic URL passing by parameter for someone who wants to use it as a playground. But IHMO is we must keep this swagger file only for we can debug the deployed node.

Pantani (commented):
We can use go swagger to auto-generate the swagger files for us in the future, which is easier to maintain and update with the code. What do you think?

https://medium.com/@pedram.esmaeeli/generate-swagger-specification-from-go-source-code-648615f7b9d9

arnaubennassar:
> We can use go swagger to auto-generate the swagger files for us in the future, which is easier to maintain and update with the code. What do you think?
> 
> https://medium.com/@pedram.esmaeeli/generate-swagger-specification-from-go-source-code-648615f7b9d9

Looks interesting! I've created an issue to discuss the adoption of this tool: https://github.com/hermeznetwork/hermez-node/issues/678

arnaubennassar:
After a short discussion with management we've decided to use the testnet api (Rinkeby) instead of the mock up server as it gives more realistic information, and it's safe in terms of nobody losing money

Pantani (approved):

api/swagger.yml
```diff
@@ -59,9 +59,9 @@ externalDocs:
   description: Find out more about Hermez network.
   url: 'https://hermez.io'
 servers:
-  - description: Hosted mock up
-    url: https://apimock.hermez.network/v1
-  - description: Localhost mock Up
+  - description: Testnet api, connected to Rinkeby deployment
+    url: https://api.testnet.hermez.io/v1
```
arnaucube, line 63: I'm not sure if this information should go here. I understand this swagger config file as the file to run the mock api locally, but not as the directory list of the current available deployed servers.
I think that Hosted mockup server should not go to this file, and should go to the documentation website (in fact, it already is there: https://docs.hermez.io/#/developers/api ).
In this way, if the public deployed servers list changes, there is no need to update this file (which is unrelated to a public list of servers, as is a swagger config file).
```suggestion
```

api/swagger.yml
```diff
@@ -59,9 +59,9 @@ externalDocs:
...
+  - description: Localhost usefull for testing/developing the api
```
arnaucube, line 64:
```suggestion
  - description: Localhost useful for testing/developing the api
```

api/swagger.yml
arnaubennassar: After an off github chat, @arnaucube made me realise that there is a major problem with using testnet for Swagger playground: there are several endpoints that are not straight forward to work in a sense of achieving a `200` response.

This produces a tradeoff situation:

- Mock server: all responses will be correct, except the ones that violate a basic validation constrains defined in Swagger, this may lead to a misunderstanding on how the validation works in a real API, but will always return interesting examples.
- "Real" testnet server: input parameters will actually produce different results and validation will work as expected. But it may be difficult or very difficult to achieve `200` respnses in some endpoints

arnaucube (approved):

PR #670, opened 2021-03-23
fix the invalid goarch build

Pantani:
### What does this MR does?

- Increase the build time avoiding call the migration-pack each build;
- Remove invalid architecture build to avoid int overlow error:
```
api/config.go:38:3: constant 4294967296 overflows int
api/config.go:46:3: constant 100000000000000 overflows int
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 bin/
+dist/
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,6 +1,7 @@
 before:
   hooks:
     - go mod download
+    - make migration-pack
 
 builds:
   - main: ./cli/node/main.go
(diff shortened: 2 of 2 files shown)


ed255:
I propose only releasing the build for linux 64 bits.  That's the platform that we have used to develop (although I think there are some new developers using MacOS), but most importantly, it's the platform that we have used in all the integration tests and long running tests.  We don't even know how the node behaves on Windows, or on arm64 arch, so I think we should only release what is officially supported.  And the community always have the chance to make a build for a non-official architecture and test by themselves without expecting official support.

Pantani:
> I propose only releasing the build for linux 64 bits. That's the platform that we have used to develop (although I think there are some new developers using MacOS), but most importantly, it's the platform that we have used in all the integration tests and long running tests. We don't even know how the node behaves on Windows, or on arm64 arch, so I think we should only release what is officially supported. And the community always have the chance to make a build for a non-official architecture and test by themselves without expecting official support.

ed255 (approved):

Mikelle (approved):

Issue #665, opened 2021-03-22
Return a better error when the API is called without version in the path


arnaubennassar:
Right now a valid call to the API just missing the version in the path returns:

```
{
    "error": "404 page not found"
}
```

Which is the default 404 response, it would be nice to have a message saying something like: `Version not provided, please provide a valid version in the path such as v1`

PR #674, opened 2021-03-24
added checker for version in no route api

Mikelle:
Close https://github.com/hermeznetwork/hermez-node/issues/665


arnaubennassar (changes requested):
I don't see when this function is called, could you add a test for this?

node/node.go
```diff
@@ -608,6 +609,13 @@ type NodeAPI struct { //nolint:golint
 }
 
 func handleNoRoute(c *gin.Context) {
+	matched, _ := regexp.MatchString(`^/v[0-9]+/`, c.Request.URL.Path)
```
ed255, line 612: The change looks good to me.  Nevertheless I think it's worth considering moving this logic of defining the NoRoute handler to the api module, and call the `engine.NoRoute` inside `api.NewAPI`.

ed255 (changes requested):

arnaubennassar (dismissed):

ed255 (approved):

api/api_test.go
```diff
@@ -41,7 +41,8 @@ type Pendinger interface {
 }
 
 const apiPort = "4010"
-const apiURL = "http://localhost:" + apiPort + "/v1/"
+const apiIP = "http://localhost:"
+const apiURL = apiIP + apiPort + "/v1/"
```
Pantani, line 45:
```suggestion
const (
	apiPort = "4010"
	apiIP   = "http://localhost:"
	apiURL  = apiIP + apiPort + "/v1/"
)
```

Pantani (approved):

OBrezhniev (approved):

Pantani (approved):

PR #677, opened 2021-03-24
Faster synchronization by fetching only block headers

OBrezhniev:
We were fetching whole Ethereum blocks where we actually needed only headers.
Fixed by switching to client.HeaderByNumber instead of BlockByNumber.
This change saves a lot of network bandwidth and CPU time.


ed255 (approved):

Pantani (dismissed):
nice improvement

jeffprestes:
LGTM

Pantani (approved):

ed255 (approved):

Issue #548, opened 2021-02-17
Endpoint to obtain the pending transactions in the PoolDB for an account


arnaucube:
This will return all the pending transactions for the given rollup account, where the transactions that have been processed by the TxSelector but not been selected in the forged batches will have the information of the reason.
And for example when a tx is still pending due not enough fee, the client (automatically or manually through the wallet UI) can get all the pending transactions and see that needs to resend the tx with same Nonce but with higher fee.

Example of endpoint:
`GET /transactions-pool/` with query parameter `idx` (not path parameter)

PR #684, opened 2021-03-26
Add greatestNonceInPool for GET /accounts

### LINKED ISSUE #681: API should give more information about nonces
Right now API only give information about the nonces having in mind the transactions that have already been forged. This may not be enough in some scenarios such as multidevice.

The `GET /accounts` endpoint should include a new field that represent the higher nonce of the pool. 

This may lead to some inconsistencies if for some reason there is gap in the nonces (examples the pool has nonces 3, 4, 7 for a given idx). We won't implement any special behaviour for this scenario as this is probably using "custom scripts" or just trying to do something nasty.

arnaubennassar:
Return the greatest nonce in the pool for each account in the GET /accounts endpoint.

This will be useful for managing  sending transactions to the pool in scenarios where a user have multiple clients, or the cache of the client gets deleted.

Close #681

### DIFF
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -2369,7 +2369,16 @@ components:
         accountIndex:
           $ref: '#/components/schemas/AccountIndex'
         nonce:
-          $ref: '#/components/schemas/Nonce'      
+          $ref: '#/components/schemas/Nonce'
+        greatestNonceInPool:
+          type: number
+          description: | 
+            The greatest nonce of among all the transactions in the account from this account. 
+            `null` indicates that there are no transactions in the pool from this account.
+            Note that there is no guarantee that all transactions of the pool from the ssame accoun have consequtive nonces,
+            but the greatest will always be returned.
+          example: 3
+          nullable: true 
         balance:
           $ref: '#/components/schemas/BigInt'
         bjj:
@@ -2385,6 +2394,7 @@ components:
         hezEthereumAddress: hez:0x7E5F4552091A69125d5DfCb7b8C2659029395Bdf
         itemId: 4
         nonce: 0
+        greatestNonceInPool: 1
         token:
           USD: 500
           decimals: 18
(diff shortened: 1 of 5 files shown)


api/swagger.yml
```diff
@@ -2369,7 +2369,16 @@ components:
...
           $ref: '#/components/schemas/AccountIndex'
         nonce:
-          $ref: '#/components/schemas/Nonce'      
+          $ref: '#/components/schemas/Nonce'
+        greatestNonceInPool:
+          type: number
+          description: | 
+            The greatest nonce of among all the transactions in the account from this account. 
+            `null` indicates that there are no transactions in the pool from this account.
+            Note that there is no guarantee that all transactions of the pool from the ssame accoun have consequtive nonces,
```
AlbertoElias, line 2378: same

AlbertoElias: account, consecutive

ed255:
I think this change breaks the separation between the `synchronizer` API side and `coordinator` API side. I think it's preferable to leave the account endpoint to just data gathered by the synchronizer, and not mix it with the tx_pool.

Also, it's not clear to me how useful this `greatest_nonce` is, as there are many situations where it's not guaranteed that a tx with `greatest_nonce`+1 will be forged.  I believe implementing this is a much better solution https://github.com/hermeznetwork/hermez-node/issues/548

arnaubennassar:
> I think this change breaks the separation between the synchronizer API side and coordinator API side. I think it's preferable to leave the account endpoint to just data gathered by the synchronizer, and not mix it with the tx_pool.

That's absolutely right! 

> Also, it's not clear to me how useful this greatest_nonce is, as there are many situations where it's not guaranteed that a tx with greatest_nonce+1 will be forged. I believe implementing this is a much better solution #548

The idea was to provide an easy solution in a multi device/browser scenario assuming the users won't do "weird stuff". With tha being said, I'm completely agree on #548 being a better solution, and will close this PR, branch and issue in favor of #548.

PR #685, opened 2021-03-29
Faster synchronization with asynchronous deletion of old checkpoints

OBrezhniev:
After each checkpoint we were doing synchronous deletion of old checkpoints, which was taking a lot of time during synchronization. 
Fixed by making DeleteOldCheckpoints public method in KVDB and StateDB, removing it from MakeCheckpoint and adding async call in TxProcessor.
Added unit tests to check that possible concurrent deletion doesn't break anything.

### DIFF
--- a/db/kvdb/kvdb.go
+++ b/db/kvdb/kvdb.go
@@ -46,12 +46,12 @@ type KVDB struct {
 	cfg Config
 	db  *pebble.Storage
 	// CurrentIdx holds the current Idx that the BatchBuilder is using
-	CurrentIdx   common.Idx
-	CurrentBatch common.BatchNum
-	m            sync.Mutex
-	mutexDelOld  sync.Mutex
-	wg           sync.WaitGroup
-	last         *Last
+	CurrentIdx      common.Idx
+	CurrentBatch    common.BatchNum
+	mutexCheckpoint sync.Mutex
+	mutexDelOld     sync.Mutex
+	wg              sync.WaitGroup
+	last            *Last
 }
 
 // Last is a consistent view to the last batch of the stateDB that can
@@ -552,8 +552,8 @@ func (k *KVDB) MakeCheckpointFromTo(fromBatchNum common.BatchNum, dest string) e
 	// synchronizer to do a reset to a batchNum at the same time as the
 	// pipeline is doing a txSelector.Reset and batchBuilder.Reset from
 	// synchronizer to the same batchNum
-	k.m.Lock()
-	defer k.m.Unlock()
+	k.mutexCheckpoint.Lock()
+	defer k.mutexCheckpoint.Unlock()
 	return PebbleMakeCheckpoint(source, dest)
 }
 
--- a/db/kvdb/kvdb.go
+++ b/db/kvdb/kvdb.go
@@ -49,6 +49,8 @@ type KVDB struct {
 	CurrentIdx   common.Idx
 	CurrentBatch common.BatchNum
 	m            sync.Mutex
+	mutexDelOld  sync.Mutex
+	wg           sync.WaitGroup
 	last         *Last
 }
 
@@ -444,10 +446,15 @@ func (k *KVDB) MakeCheckpoint() error {
 			return tracerr.Wrap(err)
 		}
 	}
-	// delete old checkpoints
-	if err := k.deleteOldCheckpoints(); err != nil {
-		return tracerr.Wrap(err)
-	}
+
+	k.wg.Add(1)
+	go func() {
+		delErr := k.DeleteOldCheckpoints()
+		if delErr != nil {
+			log.Errorw("delete old checkpoints failed", "err", delErr)
+		}
+		k.wg.Done()
+	}()
 
 	return nil
 }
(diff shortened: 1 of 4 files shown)


OBrezhniev:
@ed255, there are other places where MakeCheckpoint is called, but after these calls there's a call to TxProcessor::ProcessTxs later anyway, please verity that it's ok to call it just in one place.

txprocessor/txprocessor.go
```diff
@@ -190,6 +190,12 @@ func (tp *TxProcessor) ProcessTxs(coordIdxs []common.Idx, l1usertxs, l1coordinat
 		if err == nil {
 			err = tp.s.MakeCheckpoint()
 		}
+		go func() {
```
ed255, line 193: I think it can be dangerous to have "untracked" goroutines as a general practice.  For example, in this case you could stop the node at the middle of a `DeleteOldCheckpoints`, and then one of the old checkpoints gets half deleted, leading to possible unexpected results.  In this case I think it's not severe, because old checkpoints are not supposed to be read any time (that's why they are deleted).  But I think it's a better practice to keep track of all the go routines and make sure they are stopped on termination of the program.

I don't have the perfect solution for this because the StateDB is synchronous by nature, so it doesn't have `start` or `stop` methods, but here's an idea:
- Leave the `DeleteOldCheckpoints` inside of `MakeCheckpoint` in KVDB but run it via a goroutine.
- In KVDB add a waitgroup.  Before starting the gorotuine call `wg.Add(1)`.  And once the `DelteOldCheckpoints` is done call `wg.Done()`.
- In `KVDB.Close()` call `wg.Done()`.

ed255 (changes requested):

ed255:
> @ed255, there are other places where MakeCheckpoint is called, but after these calls there's a call to TxProcessor::ProcessTxs later anyway, please verity that it's ok to call it just in one place.

The node uses the StateDB in 3 places:
1. Synchronizer (this one, has a direct call to `ProcessTxs`)
2. Coordinator
    2.1. TxSelector (a call to `TxSelector.getL1L2TxSelection`) **I think there's no call to `ProcessTxs` here!**
    2.2. BatchBuilder (a call to `BatchBuilder.BuildBatch` which itself calls `ProcessTxs`)
    
Please, see my suggestion https://github.com/hermeznetwork/hermez-node/pull/685#discussion_r603328196 to consider a slightly different approach.

txprocessor/txprocessor.go
```diff
@@ -190,6 +190,12 @@ func (tp *TxProcessor) ProcessTxs(coordIdxs []common.Idx, l1usertxs, l1coordinat
...
+			err = tp.s.DeleteOldCheckpoints()
+			if err != nil {
+				log.Warnw("delete old checkpoints failed", "err", err)
```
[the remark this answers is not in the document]
OBrezhniev: Done

txprocessor/txprocessor.go
OBrezhniev: Good catch. Please check the updated implementation.

OBrezhniev:
>     2. Coordinator
>        2.1. TxSelector (a call to `TxSelector.getL1L2TxSelection`) **I think there's no call to `ProcessTxs` here!**
No direct call, but later result is passed to proof generator and finally to batchbuilder, which calls ProcessTx.
My idea was that maybe we can call delete old checkpoints even less frequently. That's why I wanted to move it outside MakeCheckpoint and also testing it is easier when it's outside. But after returning it back and looking at the new unit tests they kind of cover the same cases, though a bit more indirectly. Check it out, @ed255

ed255:
> > ```
> > 2. Coordinator
> >    2.1. TxSelector (a call to `TxSelector.getL1L2TxSelection`) **I think there's no call to `ProcessTxs` here!**
> > ```
> 
> No direct call, but later result is passed to proof generator and finally to batchbuilder, which calls ProcessTx.

After reading again my comment https://github.com/hermeznetwork/hermez-node/pull/685#issuecomment-809413666 I think I was not clear about the different usages of StateDB.  In particular I want to state that the 3 points I mentioned where the StateDB is used, are actually 3 independent instances of the StateDB, each one with a different KVDB.  So deleting a checkpoint in the StateDB associated to 2.2 BatchBuilder, will not touch the StateDB associated with the TxSelector.  That's why I believe that the old checkpoints in the StateDB used by the TxSelector are not being deleted with the current state of this Pull Request.

OBrezhniev:
@ed255, I haven't noticed push failed, now there are these changes I was talking about.

ed255 (approved):

db/kvdb/kvdb.go
```diff
@@ -49,6 +49,8 @@ type KVDB struct {
 	CurrentIdx   common.Idx
 	CurrentBatch common.BatchNum
 	m            sync.Mutex
+	mutexDelOld  sync.Mutex
```
Pantani, line 52: Now we have 2 mutexes. Can you please give a better name for another one? How do I know the mutex I must use?

db/kvdb/kvdb.go
```diff
@@ -49,6 +49,8 @@ type KVDB struct {
...
+	wg           sync.WaitGroup
```
Pantani, line 53:
```suggestion
	wgDelOld           sync.WaitGroup
```

db/kvdb/kvdb.go
OBrezhniev: Ok

db/kvdb/kvdb.go
OBrezhniev: This wait group is not specific to deletion, it's main idea to wait for all goroutines in kvdb on app shutdown, so in future there could be other gouroutines to wait for.

db/kvdb/kvdb.go
OBrezhniev: Done

Pantani (approved):

ed255 (approved):

Issue #680, opened 2021-03-26
Avoid creating unnecesary accounts


arnaubennassar:
Scenario:

1. Alice sends a `createAccount` tx using the wallet (L1 tx)
2. Bob sends a `transferToEthAddr` to Alice (L2 tx)
3. Coordinator forjes the L2 tx, to do so creates an account on behalf of Alice
4. Coordinator forjes the L1 tx, which also creates an account for Alice

As a result Alice has two accounts, which probably is something she didn't want. And the coordinator had to create one extra account.

There are two strategies to mitigate this:

- [x] `txselector` avoid creating new accounts on behalf of users when there are pending L1 txs that will do that anyway
- `api` expose information that helps wallet avoiding the situation:
    - [x] Pending L1 that create accounts for a given Ethereum address
    - [x] Pending L2 that may create an account on behalf of the user: will be done by #548

ed255:
I will add a tick to the first item in the checklist due to this PR being merged into develop: https://github.com/hermeznetwork/hermez-node/pull/688

jeffprestes:
Should we close this in favor of #739 or vice-versa?

arnaubennassar:
Closing this issue as all the items are covered, however this is not trivial from the point of view of a wallet, and I don't know how we should document this sort of thing

PR #686, opened 2021-03-30
Add filter to include unforged history txs in api

arnaubennassar:
- Add a query filter in `GET /transactions-history` that makes possible to include the unforged L1 user transactions in the response.

This will be useful to help clients (wallets) avoid creating unnecesary accounts, as desribed in  #680 (remind to mark the associate check as done in the issue when this is merged)

### DIFF
--- a/api/batch.go
+++ b/api/batch.go
@@ -109,7 +109,7 @@ func (a *API) getFullBatch(c *gin.Context) {
 	// Fetch txs forged in the batch from historyDB
 	maxTxsPerBatch := uint(2048) //nolint:gomnd
 	txs, _, err := a.h.GetTxsAPI(
-		nil, nil, nil, nil, batchNum, nil, nil, &maxTxsPerBatch, historydb.OrderAsc,
+		nil, nil, nil, nil, batchNum, nil, nil, nil, &maxTxsPerBatch, historydb.OrderAsc,
 	)
 	if err != nil && tracerr.Unwrap(err) != sql.ErrNoRows {
 		retSQLErr(err, c)
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -529,6 +529,15 @@ paths:
             type: integer
             minimum: 1
             maximum: 2049
+        - name: includePendingL1s
+          in: query
+          required: false
+          description: |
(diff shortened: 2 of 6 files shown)


Mikelle (dismissed):

api/batch.go
```diff
@@ -109,7 +109,7 @@ func (a *API) getFullBatch(c *gin.Context) {
 	// Fetch txs forged in the batch from historyDB
 	maxTxsPerBatch := uint(2048) //nolint:gomnd
 	txs, _, err := a.h.GetTxsAPI(
-		nil, nil, nil, nil, batchNum, nil, nil, &maxTxsPerBatch, historydb.OrderAsc,
+		nil, nil, nil, nil, batchNum, nil, nil, nil, &maxTxsPerBatch, historydb.OrderAsc,
```
ed255, line 112: This function takes many arguments now.  Please, consider now or in the future, the possibility of creating a struct with parameters to pass to the function to make it more clear which parameters are being set.  Using a struct is specially useful if some of the parameters are nil, because you can just skip setting them in the struct (and by default they will be nil)

api/swagger.yml
```diff
@@ -529,6 +529,15 @@ paths:
             type: integer
             minimum: 1
             maximum: 2049
+        - name: includePendingL1s
+          in: query
+          required: false
+          description: |
+            If set to true L1 transactions that have been addeed to the smart contract queue but haven't been forgrd yet are returned.
+            Warning: the correctness of the order is not guaranteed when using this filter, as the unforged transactions may change their position
```
ed255, line 537: Does this mean that during pagination, you can get inconsistent results if this flag is true?  For example, in the first page you get the txs as pending, and in the second pagination you get them again as forged (in the case they are forged between the first and second page query)?

ed255 (changes requested):

api/batch.go
arnaubennassar: Opened an issue to implement this in the future: https://github.com/hermeznetwork/hermez-node/issues/692

ed255 (approved):

Pantani (approved):

AlbertoElias (approved):

PR #688, opened 2021-03-31
TxSel avoid L1CoordTx for L2Tx that have a L1UserTx in the frozen queue

arnaucube:
For the L2Txs of TransferToEthAddr & TransferToBJJ for a not-yet
existing accounts, in the TxSelector check if L2Tx receiver account will
be created by a L1UserFrozenTxs (in the next batch, the current frozen
queue). In that case, the L2Tx will be discarded at the current batch,
even if there is an AccountCreationAuth for the account, as there is a
L1UserTx in the frozen queue that will create the receiver Account.  The
L2Tx is discarded to avoid the Coordinator creating a new
L1CoordinatorTx to create the receiver account, which will be also
created in the next batch from the L1UserFrozenTx, ending with the user
having 2 different accounts for the same TokenID.
The double account creation is supported by the Hermez zkRollup
specification, but it was decided to mitigate it at the TxSelector level
for the explained cases.

### DIFF
--- a/coordinator/pipeline.go
+++ b/coordinator/pipeline.go
@@ -523,15 +523,30 @@ func (p *Pipeline) forgeBatch(batchNum common.BatchNum) (batchInfo *BatchInfo,
 		if err != nil {
 			return nil, nil, tracerr.Wrap(err)
 		}
+		// l1UserFutureTxs are the l1UserTxs that are not being forged
+		// in the next batch, but that are also in the queue for the
+		// future batches
+		l1UserFutureTxs, err := p.historyDB.GetUnforgedL1UserFutureTxs(p.state.lastForgeL1TxsNum + 1)
+		if err != nil {
+			return nil, nil, tracerr.Wrap(err)
+		}
+
 		coordIdxs, auths, l1UserTxs, l1CoordTxs, poolL2Txs, discardedL2Txs, err =
-			p.txSelector.GetL1L2TxSelection(p.cfg.TxProcessorConfig, _l1UserTxs)
+			p.txSelector.GetL1L2TxSelection(p.cfg.TxProcessorConfig, _l1UserTxs, l1UserFutureTxs)
 		if err != nil {
 			return nil, nil, tracerr.Wrap(err)
 		}
 	} else {
+		// get l1UserFutureTxs which are all the l1 pending in all the
+		// queues
+		l1UserFutureTxs, err := p.historyDB.GetUnforgedL1UserFutureTxs(p.state.lastForgeL1TxsNum) //nolint:gomnd
+		if err != nil {
+			return nil, nil, tracerr.Wrap(err)
+		}
+
 		// 2b: only L2 txs
 		coordIdxs, auths, l1CoordTxs, poolL2Txs, discardedL2Txs, err =
-			p.txSelector.GetL2TxSelection(p.cfg.TxProcessorConfig)
+			p.txSelector.GetL2TxSelection(p.cfg.TxProcessorConfig, l1UserFutureTxs)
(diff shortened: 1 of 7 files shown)


coordinator/pipeline.go
```diff
@@ -523,8 +523,13 @@ func (p *Pipeline) forgeBatch(batchNum common.BatchNum) (batchInfo *BatchInfo,
 		if err != nil {
 			return nil, nil, tracerr.Wrap(err)
 		}
+		_l1UserFrozenTxs, err := p.historyDB.GetUnforgedL1UserTxs(p.state.lastForgeL1TxsNum + 2) //nolint:gomnd
```
ed255, line 526: As discussed AFK, you should get all the queues `>= p.state.lastForgeL1TxsNum + 2` instead of just that one.
Also, you should pass the pendingL1UserTxs in the `GetL2TxSelection`

txselector/txselector.go
```diff
@@ -574,18 +577,37 @@ func (txsel *TxSelector) processL2Txs(tp *txprocessor.TxProcessor,
...
+	// AccountCreationAuth for the account, as there is a L1UserTx in the
+	// frozen queue that will create the receiver Account.  The L2Tx is
+	// discarded to avoid the Coordinator creating a new L1CoordinatorTx to
+	// create the receiver account, which will be also created in the next
+	// batch from the L1UserFrozenTx, ending with the user having 2
+	// different accounts for the same TokenID. The double account creation
+	// is supported by the Hermez zkRollup specification, but it was decided
+	// to mitigate it at the TxSelector level for the explained cases.
+	if checkPendingToCreateFrozenQueue(l1UserFrozenTxs, l2Tx.TokenID, l2Tx.ToEthAddr, l2Tx.ToBJJ) {
+		return nil, nil, nil, fmt.Errorf("L2Tx discarded at the current batch," +
```
ed255, line 604: Bear in mind that this error message will appear in the Info field.  I think it's very long!

txselector/txselector.go
```diff
@@ -700,6 +722,22 @@ func checkAlreadyPendingToCreate(l1CoordinatorTxs []common.L1Tx, tokenID common.
...
 
+func checkPendingToCreateFrozenQueue(l1UserFrozenTxs []common.L1Tx, tokenID common.TokenID,
+	addr ethCommon.Address, bjj babyjub.PublicKeyComp) bool {
+	for i := 0; i < len(l1UserFrozenTxs); i++ {
+		if l1UserFrozenTxs[i].FromEthAddr == addr &&
+			l1UserFrozenTxs[i].TokenID == tokenID &&
+			l1UserFrozenTxs[i].FromBJJ == bjj {
+			return true
+		}
+		if l1UserFrozenTxs[i].FromEthAddr == addr &&
```
ed255, line 733: I don't understand this second check

ed255 (changes requested):

coordinator/pipeline.go
arnaucube: Updated, using also a new method of historydb to get the future l1usertxs that will be forged in the future batch (except the next one)

txselector/txselector.go
arnaucube: Shortened the error message. It was so long initially as this is a behavior that is not the 'default by protocol' behavior, but is a request to mitigate some particular cases, so the idea was to give as much details as possible on why the L2Tx was being discarded (even that technically with the protocol spec in the hand should be possible to be forged).
Now the `Info` message is shortened, and the details continue to be in the comment of the code, so the debugger will be able to find why the L2Tx is being discarded in those cases.

PR #696, opened 2021-04-01
fixed s.vars.Rollup.Buckets empty array

Mikelle:


synchronizer/synchronizer.go
```diff
@@ -1155,17 +1155,16 @@ func (s *Synchronizer) rollupSync(ethBlock *common.Block) (*common.RollupData, e
 	// implementation RollupEventsByBlock already inserts a non-existing
 	// RollupEventUpdateBucketsParameters into UpdateBucketsParameters with
 	// all the bucket values at 0 and SafeMode = true
-
 	for _, evt := range rollupEvents.UpdateBucketsParameters {
-		for i, bucket := range evt.ArrayBuckets {
-			s.vars.Rollup.Buckets[i] = common.BucketParams{
+		for _, bucket := range evt.ArrayBuckets {
```
ed255, line 1159: Please, add the following line before the loop:
```go
s.vars.Rollup.Buckets = []common.BucketParams{}
```
This is because otherwise, after two updates the `s.vars.Rollup.Buckets` variable will have double the items, as it is kept during all the existence of the Synchronizer object.

ed255 (changes requested):

ed255 (approved):

PR #697, opened 2021-04-02
Fix Goreleaser migration pack

Pantani:
### What does this MR does?

Fix the migration pack for the binary release distribution

## Changes

- Download the packr as Github Action
- Add Go build env vars
- Specify the `GOARCH` and `GOOS` to `packr`

**Fully tested on a forked repository:**
https://github.com/Pantani/hermez-node/releases/tag/v1.3.3

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,24 +3,25 @@ name: goreleaser
 on:
   push:
     tags:
-      - '*'
+      - 'v*.*.*'
 
 jobs:
   goreleaser:
     runs-on: ubuntu-latest
     steps:
-      -
-        name: Checkout
+      - name: Checkout
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      -
-        name: Set up Go
+      - name: Set up Go
         uses: actions/setup-go@v2
         with:
           go-version: 1.16
-      -
-        name: Run GoReleaser
+      - name: Get packr
+        run: go get -u github.com/gobuffalo/packr/v2/packr2
+      - name: Prepare
+        run: git reset --hard
+      - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v2
         with:
           version: latest
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,7 +1,11 @@
+env:
+  - GO111MODULE=on
+  - CGO_ENABLED=0
+
 before:
   hooks:
+    - go mod tidy
     - go mod download
-    - make migration-pack
 
 builds:
   - main: ./cli/node/main.go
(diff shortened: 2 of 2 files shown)


OBrezhniev (approved):
LGTM

.github/workflows/release.yml
```diff
@@ -3,27 +3,28 @@ name: goreleaser
...
+      - name: Prepare
+        run: git reset --hard
+      - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v2
         with:
           version: latest
           args: release --rm-dist
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
ed255, line 30: now that the repo is public, is this still necessary?

.goreleaser.yml
```diff
@@ -1,7 +1,11 @@
+env:
+  - GO111MODULE=on
+  - CGO_ENABLED=0
```
ed255, line 3: Could you explain why you disable `CGO`?

.goreleaser.yml
```diff
@@ -12,14 +16,13 @@ builds:
...
-archives:
-  - replacements:
-      darwin: Darwin
-      linux: Linux
-      windows: Windows
-      386: i386
-      amd64: x86_64
+    hooks:
+      pre:
+        - cmd: bash -c "GOOS={{ .Os }} GOARCH={{ .Arch }} packr2"
```
ed255, line 21: The last time I tried using packr, it only worked if run in the folder where packr is used (https://github.com/hermeznetwork/hermez-node/blob/d5d2812cab0dd8c8eec3e65c8e3e0dc09f96cb29/db/utils.go#L34)
I believe this will run packr in the root of the project.  Have you checked moving this binary to another system and running it?

ed255 (changes requested):

.github/workflows/release.yml
Pantani: we need it for we can publish the binary into the releases page

.goreleaser.yml
Pantani: the folder was specified in the next line:

https://github.com/hermeznetwork/hermez-node/pull/697/files#diff-42e26dc67aed8aa3edb2472b4403288c1699fb6dc47419b9a475f0f224fe4689R22

.goreleaser.yml
Pantani: I remove it because we are not using

ed255: thanks for the reply! I want to mention that [pebble](https://github.com/cockroachdb/pebble) has different behavior internally depending on whether CGO is enabled or not.  I know this because when we started the hermez-node we used a version of pebble that had a bug when CGO was disabled!  The bug was fixed in November 2020.  See here for mentions of CGO in pebble: https://github.com/cockroachdb/pebble/search?q=CGO&type=issues

I'm mentioning this to be aware that one of our dependencies is "sensitive" to CGO, to consider this in the future if any issue occurs!

.goreleaser.yml
ed255: Ah! I missed that, thanks!

ed255 (approved):

Issue #682, opened 2021-03-26
Add info on tx_pool for invalid transactions


arnaubennassar:
The `tx_pool` sql table has a column `info` which is really useful for debugging and in the future could even be used for UIs. 

As of now, this column is only set for pending transactions that have not been selected, for instance due to a low fee, not enough balance, ...

However this is not set for invalid transactions. It will be really useful to also add this column in this case.

PR #698, opened 2021-04-05
added info if transaction is invalid

Mikelle:
Close https://github.com/hermeznetwork/hermez-node/issues/682
Found only one place, when transacting become invalid - InvalidateOldNonces method, so only added info there.

### DIFF
--- a/db/l2db/l2db.go
+++ b/db/l2db/l2db.go
@@ -319,7 +319,7 @@ func (l2db *L2DB) GetPendingUniqueFromIdxs() ([]common.Idx, error) {
 	return idxs, nil
 }
 
-const invalidateOldNoncesInfo = `Nonce is smaller or equal than account nonce`
+const invalidateOldNoncesInfo = `Nonce is smaller than account nonce`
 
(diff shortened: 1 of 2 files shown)


Pantani (dismissed):

db/l2db/l2db.go
```diff
@@ -319,9 +319,12 @@ func (l2db *L2DB) GetPendingUniqueFromIdxs() ([]common.Idx, error) {
 	return idxs, nil
 }
 
+const invalidateOldNoncesInfo = `Nonce is smaller or equal than account nonce`
```
ed255, line 322: I think it should just be `smaller than`. See the SQL query: `tx_pool.nonce < updated_acc.nonce;`

ed255 (approved):

db/l2db/l2db.go
ed255, line 322: I think it should just be `smaller than`. See the SQL query: `tx_pool.nonce < updated_acc.nonce;`

ed255 (changes requested):

ed255 (approved):

Pantani (approved):

Issue #692, opened 2021-03-31
Consider using struct in functions that take too many arguments


arnaubennassar:
Specially some of the functions from `apiqueries.go` takes a lot of arguments. In order to keep the code clean and redable, it would be nice to use structs to group the input parameters as @ed255 pointed.

PR #699, opened 2021-04-06
Refactor/apiqueries structs

Mikelle:
Close https://github.com/hermeznetwork/hermez-node/issues/692.
Adding structs to group the input parameters.
Use github squash and merge, if possible, when you will merge the PR


api/requests/get_accounts.go
```diff
@@ -0,0 +1,18 @@
+package requests
+
+import (
+	ethCommon "github.com/ethereum/go-ethereum/common"
+	"github.com/hermeznetwork/hermez-node/common"
+	"github.com/iden3/go-iden3-crypto/babyjub"
+)
+
+// GetAccountsAPIRequest is an API request struct for getting accounts
+type GetAccountsAPIRequest struct {
```
ed255, line 10: I see that you created a new file for each `Request` type.  I think this introduces too many files and makes it a bit harder to navigate the code.  Here are my suggestions (what I would have done differently):
1. I would declare the types in the same file as the function that uses them.  For each API function, just before it declare the `Request` type used for that function.  I don't understand the need to split them into files, and I don't understand the need to move them to another package.

api/requests/get_batches.go
```diff
@@ -0,0 +1,15 @@
+package requests
+
+import ethCommon "github.com/ethereum/go-ethereum/common"
+
+// GetBatchesAPIRequest is an API request struct for getting batches
+type GetBatchesAPIRequest struct {
```
ed255, line 6: 2. This is a subjective comment: I would have called these structs `...Args` (instead of `...Request`) because they are the arguments to a function, and then inside the function, I would have called the variable `args` (instead of `request`).  This is very subjective so feel free to ignore it.

api/account.go
```diff
@@ -37,8 +38,16 @@ func (a *API) getAccounts(c *gin.Context) {
...
+		TokenIDs: tokenIDs,
+		EthAddr:  addr,
+		Bjj:      bjj,
+		FromItem: fromItem,
+		Limit:    limit,
+		Order:    order,
+	}
 	// Fetch Accounts from historyDB
-	apiAccounts, pendingItems, err := a.h.GetAccountsAPI(tokenIDs, addr, bjj, fromItem, limit, order)
+	apiAccounts, pendingItems, err := a.h.GetAccountsAPI(request)
```
ed255, line 50: This is a subjective comment.  Have you considered calling the method without declaring the request variable?

Like this:
```go
	apiAccounts, pendingItems, err := a.h.GetAccountsAPI(requests.GetAccountsAPIRequest{
		TokenIDs: tokenIDs,
		EthAddr:  addr,
		Bjj:      bjj,
		FromItem: fromItem,
		Limit:    limit,
		Order:    order,
	})
```
It looks cleaner to me!

ed255 (changes requested):

api/requests/get_accounts.go
jeffprestes: I agree. 
I don't think we would need to create another package and I also Go source files standard don't use _ between the names, except for the test files. 
I would suggest to include all the structs into parameter.go file and within same package.

Mikelle: @jeffprestes and @ed255 thank you for the suggestions! But what I should do, if placing those structs in the same package will cause cyclical dependence?

ed255 (approved):

jeffprestes (approved):

Pantani (approved):

PR #707, opened 2021-04-07
add github templates and automations

Pantani:
### What does this MR does?

Create Github template files for better organization and automation.

### Changes

- Creates templates into `.github` folder

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,24 @@
+# Contributing
+
+Thank you for considering to help out with the source code! We welcome 
+contributions from anyone on the internet, and are grateful for even the 
+smallest of fixes!
+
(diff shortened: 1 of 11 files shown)


jeffprestes (approved):

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,24 @@
+# Contributing
```
ed255, line 1: Maybe add information about the CLA (Contributor License Agreement) that is required to contribute? (If we still require it, that is)

ed255 (changes requested):

.github/CONTRIBUTING.md
jeffprestes: That's right. But how could we describe it here? Could you state a suggestion here @ed255 ?

Issue #705, opened 2021-04-07
Inconsistent case in toBjj parameter in API


AlbertoElias:
## Summary of Bug

/tx-pool has parameters called `fromBJJ` and `toBjj`, while /tx-history has both paramters in upper case `fromBJJ` and `toBJJ`.
### Expected Behavior

They all have the same case. Either BJJ or Bjj
### Steps to Reproduce

https://apidoc.hermez.network/#/Coordinator/getPoolTx
https://apidoc.hermez.network/#/Explorer/getHistoryTx

PR #709, opened 2021-04-08
changed toBjj -> toBJJ in tx-pool get request

Mikelle:
Close https://github.com/hermeznetwork/hermez-node/issues/705

### DIFF
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -1306,7 +1306,7 @@ components:
           pattern: "^hez:0x[a-fA-F0-9]{40}$"
           example: null
           nullable: true
-        toBjj:
+        toBJJ:
           type: string
           description: "BabyJubJub compressed public key, encoded as base64 URL (RFC 4648), which result in 33 bytes. The padding byte is replaced by a sum of the encoded bytes."
           pattern: "^hez:[A-Za-z0-9_-]{44}$"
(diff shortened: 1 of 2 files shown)


jeffprestes (approved):

AlbertoElias (approved):

api/swagger.yml
```diff
@@ -1410,7 +1410,7 @@ components:
         state: pend
```
ed255, line 1410: Although it was not mentioned in the issue, the type `PostPoolL2Transaction` has the field `toBjj` https://github.com/hermeznetwork/hermez-node/blob/569150a6ccba235235a41348fa0aa03d583c1713/api/swagger.yml#L1189
This is an input type.  I think it makes sense to replace all instances of `Bjj` to `BJJ` in the API.  @AlbertoElias are you OK with more changes of `Bjj -> BJJ` than the one you reported?

ed255 (changes requested):

api/swagger.yml
AlbertoElias: I'm fine. Though we do need to keep in mind that this is a breaking change unless you continue to support current naming until we move to a /v2

jeffprestes: Despite hermez-integration examples created by @Pantani and our frontend, are we aware of any other project is using our API so far? Because within project I believe we can manage this change.
However, because as we're live I don't think would be good to have a break changes and I would skip this change and create another issue to implement it in a new API version.

AlbertoElias: I think this is small enough to be okay to wait for a new API version. But I do think there could be other changes that might need to be applied faster and would require some temporary backward compatible fixes.

jeffprestes: @AlbertoElias feel free to open another issue to describe these other changes you think might be necessaries.

jeffprestes: So, @ed255 I believe it would be better move forward this PR way it is and let this item you have pointed out in another task for another release.

Pantani (approved):

ed255 (approved):

PR #712, opened 2021-04-08
Fix data race in kvdb unit tests

OBrezhniev:
There were data races in unit tests of kvdb. This PR fixes them.

### DIFF
--- a/db/kvdb/kvdb_test.go
+++ b/db/kvdb/kvdb_test.go
@@ -220,8 +220,8 @@ func TestConcurrentDeleteOldCheckpoints(t *testing.T) {
 		err = db.MakeCheckpoint()
 		require.NoError(t, err)
 		go func() {
-			err = db.DeleteOldCheckpoints()
-			require.NoError(t, err)
+			errDelete := db.DeleteOldCheckpoints()
+			require.NoError(t, errDelete)
 			wg.Done()
 		}()
 	}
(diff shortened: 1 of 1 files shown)


db/kvdb/kvdb_test.go
```diff
@@ -220,8 +220,8 @@ func TestConcurrentDeleteOldCheckpoints(t *testing.T) {
 		err = db.MakeCheckpoint()
 		require.NoError(t, err)
 		go func() {
-			err = db.DeleteOldCheckpoints()
-			require.NoError(t, err)
+			err2 := db.DeleteOldCheckpoints()
```
Pantani, line 223: try to use a more intuitive name
```suggestion
			errDelete := db.DeleteOldCheckpoints()
```

Pantani (changes requested):

stunndard (approved):
I wonder how it passed github tests. No `-race` specified there?

Pantani:
> I wonder how it passed github tests. No `-race` specified there?

No, this tests runs without the `-race` flag. But I create this PR to add it

https://github.com/hermeznetwork/hermez-node/pull/713

db/kvdb/kvdb_test.go
Pantani: also, only adding the `:` works too, because you are going to assign a new variable for this small scope

 `err := db.DeleteOldCheckpoints()`

Pantani (approved):

PR #714, opened 2021-04-08
remove max tx for remove by TTL

### LINKED ISSUE #701: Non forjable old TXs do not get deleted from txPool after TTL
TTL only applies if the pool is full. It should be appliead always.

Pantani:
### What does this MR does?

  Remove pending transactions that exceed TTL unconditionally, without the need to reach a max number of transactions.
 
## Changes

- Remove the length of the maximum transaction from the query to purge the pending transactions
 
 closes #701

### DIFF
--- a/db/l2db/l2db.go
+++ b/db/l2db/l2db.go
@@ -375,14 +375,12 @@ func (l2db *L2DB) Purge(currentBatchNum common.BatchNum) (err error) {
 		`DELETE FROM tx_pool WHERE (
 			batch_num < $1 AND (state = $2 OR state = $3)
 		) OR (
-			(SELECT count(*) FROM tx_pool WHERE state = $4) > $5 
-			AND timestamp < $6 AND state = $4
+			state = $4 AND timestamp < $5
 		);`,
 		currentBatchNum-l2db.safetyPeriod,
 		common.PoolL2TxStateForged,
 		common.PoolL2TxStateInvalid,
 		common.PoolL2TxStatePending,
-		l2db.maxTxs,
 		time.Unix(now-int64(l2db.ttl.Seconds()), 0),
 	)
 	return tracerr.Wrap(err)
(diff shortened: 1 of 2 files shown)


jeffprestes:
I'd recommend to ask to @krlosMata if he could review it in order to check whether the logic is compliant to the protocol.

krlosMata:
Transactions pool behavior does not involve `hermez-protocol` specification since any `hermez-node` could implement their own logic and setting their own behavior in order to optimize, select and delete L2 transactions xD
In any case, my thought here is that it is good to remove transactions form the pool without waiting until is full in order to avoid having a bunch of old transactions stored that do not comply our transaction pool requirements.

ed255 (approved):

db/l2db/l2db.go
```diff
@@ -375,14 +375,12 @@ func (l2db *L2DB) Purge(currentBatchNum common.BatchNum) (err error) {
 		`DELETE FROM tx_pool WHERE (
 			batch_num < $1 AND (state = $2 OR state = $3)
 		) OR (
-			(SELECT count(*) FROM tx_pool WHERE state = $4) > $5 
-			AND timestamp < $6 AND state = $4
+			state = $4 AND timestamp < $5
```
ed255, line 378: As mentioned previously in a meeting, it would be good to update the config documentation regarding this change: 
https://github.com/hermeznetwork/hermez-node/blob/c25d8ca95c5e6cd35d66b2da35df18d76de410e1/config/config.go#L152

ed255 (changes requested):

jeffprestes (approved):

ed255 (approved):

jeffprestes:
I am closing this PR due the fix was made in #720 .
If needed we can reopen it.

jeffprestes:
Sorry. It is related to another issue. Reopening the case.

Issue #718, opened 2021-04-12
L2Txs in forged state when batch fails but another node forges a batch with that batchnum


ed255:
## Summary of Bug

When there are multiple coordinators the following can happen:
- Boot Coordinator starts forging a tx into a batch with BatchNum=N, and sends the batch to ethereum.
- Coordinator A forges an empty batch with BatchNum=N, and sends batch to ethereum.
- Batch from Boot Coordinator fails, batch from Coordinator A suceeds
- Boot Coordinator ends up with the tx marked as forged in the pool, but this tx was not forged in any "final" batch.

### Expected Behavior

If the batch forged by a coordinator is discarded, all the state changes in the pool regarding that batch should be discarded, even if there's another batch with the same batchNum by another coordinator

### Steps to Reproduce

Register an alternative coordinator and achieve the following:
- Both coordinators try to forge a batch with BatchNum=N at the same time, and only one suceeds.  The one that fails should have attempted to add an l2tx that the one that succeeds didn't have.

### Additional Information:

This bug affects the multi coordinator funcionality

PR #719, opened 2021-04-12
Reset pool txs with batchNum forged externally

ed255:
Resolve https://github.com/hermeznetwork/hermez-node/issues/718

### DIFF
--- a/coordinator/coordinator.go
+++ b/coordinator/coordinator.go
@@ -490,6 +490,33 @@ func (c *Coordinator) handleMsgSyncBlock(ctx context.Context, msg *MsgSyncBlock)
 	if c.pipeline != nil {
 		c.pipeline.SetSyncStatsVars(ctx, &msg.Stats, &msg.Vars)
 	}
+
+	// If there's any batch not forged by us, make sure we don't keep
+	// "phantom forged l2txs" in the pool.  That is, l2txs that we
+	// attempted to forge in BatchNum=N, where the forgeBatch transaction
+	// failed, but another batch with BatchNum=N was forged by another
+	// coordinator successfully.
+	externalBatchNums := []common.BatchNum{}
+	for _, batch := range msg.Batches {
+		if batch.Batch.ForgerAddr != c.cfg.ForgerAddress {
+			externalBatchNums = append(externalBatchNums, batch.Batch.BatchNum)
+		}
+	}
+	if len(externalBatchNums) > 0 {
+		// If we just synced external batches, make sure the pipeline
+		// is stopped
+		if c.pipeline != nil {
+			if err := c.handleStopPipeline(ctx, "synced external batches", 0); err != nil {
+				return err
+			}
+		}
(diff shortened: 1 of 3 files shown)


ed255:
@Pantani This issue hopefully fixes an issue encountered when testing a multi-coordinator setup.  I think it would be ideal if we can wait until this change is tested in internal testnet before merging this.

jeffprestes:
Do you mean merge in master @ed255 ?

ed255:
> Do you mean merge in master @ed255 ?
> If don't, what would be the problem to merge in Develop?

I originally meant merge in develop.  After chatting with @jeffprestes he mentioned that it's better to follow the general flow of changes before testing in integration-testing:
1. Make a PR against develop
2. Someone review the PR and finally merge the PR into develop
3. Now the integration-tests run on develop with the change
4. Test develop branch / test tags from develop branch  in integration-testing
And I agree, so @Pantani please ignore my previous comment to wait, and review this PR whenever you are available, so that it can be tested in integration-testing once it's in the develop branch.

Pantani (approved):

coordinator/coordinator.go
```diff
@@ -490,6 +490,33 @@ func (c *Coordinator) handleMsgSyncBlock(ctx context.Context, msg *MsgSyncBlock)
...
+	for _, batch := range msg.Batches {
+		if batch.Batch.ForgerAddr != c.cfg.ForgerAddress {
+			externalBatchNums = append(externalBatchNums, batch.Batch.BatchNum)
+		}
+	}
+	if len(externalBatchNums) > 0 {
+		// If we just synced external batches, make sure the pipeline
+		// is stopped
+		if c.pipeline != nil {
+			if err := c.handleStopPipeline(ctx, "synced external batches", 0); err != nil {
```
Pantani, line 509: I have a doubt. Suppose we have 3 external batches from another coordinator and call the `c.handleStopPipeline` without the `failedBatchNum` parameter. It's going to reorg only the last batch, not the last 3 batches. Is this can cause a bug?

https://github.com/hermeznetwork/hermez-node/blob/0e6fc9aa78b711aee0d6bd52cd1c5e4da029b96f/coordinator/coordinator.go#L557-L572

ed255: Notice that after the call to `handleStopPipeline`, for each external batch there's a call to `l2DB.ResetForged`, which sets all the l2txs with a corresponding batchNum to pending, which is similar to what `l2DB.Reorg` does.

After reading your comment, maybe there's a more clear way to implement this without introducing a new function `l2DB.ResetForged`...

ed255: I've made an alternative implementation of the same fix here https://github.com/hermeznetwork/hermez-node/pull/720 please, take a look

Pantani: Can we call the reorg only one time passing only the lower block, do we need reorg? Is this can be safer?

PR #720, opened 2021-04-13
Reset pool txs with batchNum forged externally (alternative)

### LINKED ISSUE #718: L2Txs in forged state when batch fails but another node forges a batch with that batchnum

ed255:
Alternative implementation of https://github.com/hermeznetwork/hermez-node/pull/719
Resolve #718

### DIFF
--- a/coordinator/coordinator.go
+++ b/coordinator/coordinator.go
@@ -490,6 +490,34 @@ func (c *Coordinator) handleMsgSyncBlock(ctx context.Context, msg *MsgSyncBlock)
 	if c.pipeline != nil {
 		c.pipeline.SetSyncStatsVars(ctx, &msg.Stats, &msg.Vars)
 	}
+
+	// If there's any batch not forged by us, make sure we don't keep
+	// "phantom forged l2txs" in the pool.  That is, l2txs that we
+	// attempted to forge in BatchNum=N, where the forgeBatch transaction
+	// failed, but another batch with BatchNum=N was forged by another
+	// coordinator successfully.
+	externalBatchNums := []common.BatchNum{}
+	for _, batch := range msg.Batches {
+		if batch.Batch.ForgerAddr != c.cfg.ForgerAddress {
+			externalBatchNums = append(externalBatchNums, batch.Batch.BatchNum)
+		}
+	}
+	if len(externalBatchNums) > 0 {
+		// If we just synced external batches, make sure the pipeline
+		// is stopped
+		lastValidBatch := externalBatchNums[0] - 1
+		if c.pipeline != nil {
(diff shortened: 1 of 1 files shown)


coordinator/coordinator.go
```diff
@@ -490,6 +490,34 @@ func (c *Coordinator) handleMsgSyncBlock(ctx context.Context, msg *MsgSyncBlock)
...
+	externalBatchNums := []common.BatchNum{}
+	for _, batch := range msg.Batches {
+		if batch.Batch.ForgerAddr != c.cfg.ForgerAddress {
+			externalBatchNums = append(externalBatchNums, batch.Batch.BatchNum)
+		}
+	}
+	if len(externalBatchNums) > 0 {
+		// If we just synced external batches, make sure the pipeline
+		// is stopped
+		lastValidBatch := externalBatchNums[0] - 1
```
Pantani, line 508: Is this the last batch or only the last item of the slice? It's sorted by batch?

Pantani: I prefer this implementation, but Is it not better to use the lower batch for the reorg? I don't know how is this array is sorted

ed255: The batches in a block (which is the slice passed in `msg.Batches`) always appear sequentially, so the first one is the smallest one in the slice.
So Imagine an external coordinator forged batches 10, 11, 12. In `lastValidBatch` we store 9, and when doing the l2DB.Reorg(9), we are setting all the txs that we marked as non-pending (forging, invalid, forged} associated with batches {10, 11, 12, 13, etc...} back to pending pending.  Because we know that everything that we marked with a batchNum that has been forged externally will not end up in the real batch for that BatchNum, so we reset all the states we changed in those phantom batches we calculated.

Pantani (approved):

PR #733, opened 2021-04-19
allow overriding test DB details using env vars (+ fix #716)

### LINKED ISSUE #716: Standarlize test db credentials
## Summary of Bug

In different parts of the project there are tests that use different db credentials. Some are hardcoded. 
As the project has a make task that starts a postgres db for tests purposes, it would be useful to use the same credentials. It could be defined by reading them from config file and/or use the system variables defined at makefile. 

One example of this error is when you run API tests:

```bash 
make test-api-server
  >  Running unit tests
panic: pq: password authentication failed for user "hermez"

goroutine 1 [running]:
github.com/hermeznetwork/hermez-node/api.TestMain(0xc003af3100)
        /Users/jeffprestes/projetos/go/src/github.com/hermeznetwork/hermez-node/api/api_test.go:209 +0x6907
main.main()
        _testmain.go:117 +0x272
FAIL    github.com/hermeznetwork/hermez-node/api        16.783s
FAIL
make: *** [test-api-server] Error 1
``` 

### Expected Behavior

No error messages being displayed.

### Steps to Reproduce

```bash
make run-database-container
make test-api-server
```
### System information

N/A

### Additional Information:

It would be important to scan other test files in other packages not only in the *api* one because the issue might happen in other packages too.

stunndard:


db/utils.go
```diff
@@ -97,6 +99,31 @@ func InitSQLDB(port int, host, user, password, name string) (*sqlx.DB, error) {
 	return db, nil
 }
 
+// InitTestSQLDB opens test PostgreSQL database
```
jeffprestes, line 102: I am afraid if we have another method to connect to the database it can break the idea of unit test because the tests will not use the same init database connection. What about to make the just one point of database connection with values comming from system environment?

.github/workflows/test.yml
```diff
@@ -20,6 +20,6 @@ jobs:
       run: docker run --rm --name hermez-db-test -p 5432:5432 -e POSTGRES_DB=hermez -e POSTGRES_USER=hermez -e POSTGRES_PASSWORD="${{ secrets.POSTGRES_PASS }}" -d postgres
     - name: Test
       env:
-        POSTGRES_PASS: ${{ secrets.POSTGRES_PASS }}
+        PGPASSWORD: ${{ secrets.POSTGRES_PASS }}
```
Pantani, line 23: Did you test it? Does it work?

This env vars is for the Postgres containers. Maybe it can break the test workflow. We are using `POSTGRES PASS` in the docker commands. Check line 20:

https://github.com/hermeznetwork/hermez-node/blob/eff2955a0490d837e56761cbf551e06626f6f622/.github/workflows/test.yml#L20

ed255 (approved):

Pantani:
please, don't forget to link the issue to the PR to be easier to track and close them automatically

https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue

db/utils.go
stunndard: Please look at
https://github.com/hermeznetwork/hermez-node/pull/733/files#diff-82bdf3f3f39ca350a5ba8e048300773b62463fa65153367662399b11b9669059R124

InitSQLDB is called as before.
I made a new function only to avoid code bloat because there are too many places where the test database is opened.

jeffprestes: I see. It's only a wrapper to the regular InitSQLDB method. Thanks for the clarification.

db/utils.go
```diff
@@ -97,6 +99,31 @@ func InitSQLDB(port int, host, user, password, name string) (*sqlx.DB, error) {
...
+		user = "hermez"
+	}
+	pass := os.Getenv("PGPASSWORD")
+	if pass == "" {
+		panic("No PGPASSWORD envvar specified")
+	}
+	dbname := os.Getenv("PGDATABASE")
+	if dbname == "" {
+		dbname = "hermez"
+	}
```
Pantani, line 123: There are the redundancy default variables in the makefile. Maybe make sense we use it only in one place

https://github.com/hermeznetwork/hermez-node/pull/733/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R23-R28

jeffprestes (approved):

.github/workflows/test.yml
stunndard: The tests seem to have passed successfully. 
 https://github.com/hermeznetwork/hermez-node/runs/2378696525?check_suite_focus=true

Both make test-unit and make test-api-server work locally as well.

stunndard: POSTGRES_PASS was used only in tests, docker container uses -e POSTGRES_PASSWORD= on the command line.

I decided to rename env var because:
 - new variables are added. The POSTGRES_ prefix is too long, it can be annoying if you need to specify multiple variables on the command line;
 - these variable names are well known and understood by most of Postgres related software (at least all software using libpq)
https://www.postgresql.org/docs/current/libpq-envars.html

Pantani: You are setting the password 2 times, first, one in the command, using the var `POSTGRES_PASSWORD`: 

```
run: docker run --rm --name hermez-db-test -p 5432:5432 -e POSTGRES_DB=hermez -e POSTGRES_USER=hermez -e POSTGRES_PASSWORD="${{ secrets.POSTGRES_PASS }}" -d postgres
```

And the second one by another env var `PGDATABASE`

Pantani: And this is not related to the local unit tests. The `make` command goes to work. You must test it against a GitHub action pipeline. GitHub uses the workflow folder to creates the CI pipeline

Pantani: also, because we are using it `-e POSTGRES_PASSWORD="${{ secrets.POSTGRES_PASS }}"`  we don't need this env var, or we can do `POSTGRES_PASSWORD="$(POSTGRES_PASSWORD)"` and use it

Pantani: or if do you prefer to use the new Postgres variables you can change it to `-e PGPASSWORD="${{ secrets.POSTGRES_PASS }}` along if the others in the command

stunndard: > first, one in the command, using the var `POSTGRES_PASSWORD`:

This variable is for the docker container
https://hub.docker.com/_/postgres
It only exists inside the container

> And the second one by another env var `PGDATABASE`

This variable is used by `go test` (to connect to the docker container)

db/utils.go
stunndard: I tried to make it easy to use with and without Makefile.

.github/workflows/test.yml
Pantani: yes, sure, but we are not using this reference, because inside the container was getting the variables directly from the github secret action not from this another var:

`-e POSTGRES_PASSWORD="${{ secrets.POSTGRES_PASS }}" `

db/utils.go
jeffprestes: IMHO, this part should remain.

.github/workflows/test.yml
Pantani: Oh, sorry, I understand now. It's confusing because you are using the same name the database container uses. The context is different. If we have another database vars, the more extensive name, it's clearer and easier to understand, and we must follow all other vars standards.

db/utils.go
Pantani: If you are doing it by code, you already cover both situations. The Makefile is redundant. I think only one check is necessary.

stunndard: We need a port, password, username, and so on in the Makefile for the docker container too.
Only PGHOST is really redundant here.

Of course, we can hardcode all docker parameters, but I like things like this:
```
export PGPORT=15432
make run-database-container
make test-unit
make stop-database-container
```
It allows us to start temporary postgresql instance on a nonstandard port and do tests with it. This could be useful if you have main postgresql instance on the 5432 port and  want to avoid any data changes in it.

Pantani: It's up to you because it's a suggestion, but less code is better for readability and maintenance. If we change the default variables, we need to remember to change it into the code and also makefile... we must remember it

jeffprestes: @stunndard I think it's a good idea not depend all time of Makefile or externally change some variable just for one specific test that you will not use anymore and clearly you *do not want to change the default variables*

Pantani: and also, it's a good to try avoid hardcoded variables

Issue #710, opened 2021-04-08
Improve config file validation


OBrezhniev:
## Rationale
Right now validation of values is done by custom code in the node initialization code and in different other places around the code and there are inconsistencies, in some cases it throws errors:
https://github.com/hermeznetwork/hermez-node/blob/e38413db4463588a69ea6de8c9d71ce702b0797a/node/node.go#L226-L230

in others - just logs a message:
https://github.com/hermeznetwork/hermez-node/blob/e38413db4463588a69ea6de8c9d71ce702b0797a/priceupdater/priceupdater.go#L176-L179

Also it would be good to have default values for config params (where it makes sense) and make them optional to keep the file small and clean.

This would also improve user experience by requiring less changes to config file from the user on node updates, because right now on each addition/removal/name change of config parameter in the code the user is required to update the config file.

## Implementation
Validation rules can be added uniformly in config structure definition and validated on node start. 
See https://github.com/go-playground/validator (that is already used to check "required" fields) for available rules.
Only more complicated checks would require custom code, like for PriceUpdater.

jeffprestes:
Thanks @OBrezhniev . In order to help who's going to take this task, could you please share in example of the situation you mentioned?

OBrezhniev:
Sure, updated the issue description, @jeffprestes

PR #734, opened 2021-04-19
changed validation of config

Mikelle:
Close https://github.com/hermeznetwork/hermez-node/issues/710


config/config.go
```diff
@@ -167,27 +167,27 @@ type Coordinator struct {
...
 		// InvalidateBlockDelay is the delay between blocks to mark
 		// invalid transactions due to nonce lower than the account
 		// nonce.
-		InvalidateBlockDelay int64 `validate:"required"`
+		InvalidateBlockDelay int64 `validate:"required,gte=0"`
 	} `validate:"required"`
 	TxSelector struct {
 		// Path where the TxSelector StateDB is stored
-		Path string `validate:"required"`
+		Path string `validate:"required,file"`
```
ed255, line 178: The `file` requirement seems to check that the string is an existing file: https://github.com/go-playground/validator/blob/b926bf0df97108effd04f9325365cd7969935e16/baked_in.go#L1289
But in this case we have a different requirement
1. It's not a file, it's a path
2. The current implementation doesn't require the path to exists, it will create it if it doesn't exists.  The real requirement would be that the path can be created and written into by the user running the node.  The same applies to the rest of the fields called `Path` that have the constraint `file`.

config/config.go
```diff
@@ -196,13 +196,13 @@ type Coordinator struct {
...
 		// the ethereum node
-		GasPriceIncPerc int64
+		GasPriceIncPerc int64 `validate:"gte=0"`
 		// CheckLoopInterval is the waiting interval between receipt
 		// checks of ethereum transactions in the TxManager
 		CheckLoopInterval Duration `validate:"required"`
 		// Attempts is the number of attempts to do an eth client RPC
 		// call before giving up
-		Attempts int `validate:"required"`
+		Attempts int `validate:"required,gt=1"`
```
ed255, line 205: I think `Attempts=1` should be valid, so it would be `gte=1`

config/config.go
```diff
@@ -284,25 +284,25 @@ type Node struct {
...
 		// TokensConfig to specify how each token get it's price updated
 		TokensConfig []priceupdater.TokenConfig
 	} `validate:"required"`
 	StateDB struct {
 		// Path where the synchronizer StateDB is stored
-		Path string `validate:"required"`
+		Path string `validate:"required,file"`
 		// Keep is the number of checkpoints to keep
-		Keep int `validate:"required"`
+		Keep int `validate:"required,lt=128"`
```
ed255, line 299: I don't think we should impose an upper bound here.

config/config.go
```diff
@@ -334,12 +334,12 @@ type Node struct {
 		Explorer bool
 		// UpdateMetricsInterval is the interval between updates of the
 		// API metrics
-		UpdateMetricsInterval Duration
+		UpdateMetricsInterval Duration `validate:"required_with=Address"`
```
ed255, line 337: I don't understand the `"required_with=Address"` requirement.  This field is mandatory no matter if `Address` is set or not.  Same for `UpdateRecommendedFeeInterval`.  What does `"required_with=Address"` do?

config/config.go
```diff
@@ -334,12 +334,12 @@ type Node struct {
...
 		// UpdateRecommendedFeeInterval is the interval between updates of the
 		// recommended fees
-		UpdateRecommendedFeeInterval Duration
+		UpdateRecommendedFeeInterval Duration `validate:"required_with=Address"`
 		// Maximum concurrent connections allowed between API and SQL
-		MaxSQLConnections int `validate:"required"`
+		MaxSQLConnections int `validate:"required,gte=0"`
```
ed255, line 342: I suggest setting `gte=1`, because otherwise the API will not work

config/config.go
```diff
@@ -358,7 +358,7 @@ type APIServer struct {
 		// Explorer enables the Explorer API endpoints
 		Explorer bool
 		// Maximum concurrent connections allowed between API and SQL
-		MaxSQLConnections int `validate:"required"`
+		MaxSQLConnections int `validate:"required,gte=0"`
```
ed255, line 361: I suggest setting gte=1, because otherwise the API will not work

config/config.go
```diff
@@ -374,15 +374,15 @@ type APIServer struct {
 			// stored in the pool.  Once this number of pending L2Txs is
 			// reached, inserts to the pool will be denied until some of
 			// the pending txs are forged.
-			MaxTxs uint32 `validate:"required"`
+			MaxTxs uint32 `validate:"required,gte=0"`
```
ed255, line 377: Since this is a `uint32`, I think `gte=0` is not doing anything.

config/config.go
```diff
@@ -70,13 +70,13 @@ type Coordinator struct {
 	} `validate:"required"`
 	// ConfirmBlocks is the number of confirmation blocks to wait for sent
 	// ethereum transactions before forgetting about them
-	ConfirmBlocks int64 `validate:"required"`
+	ConfirmBlocks int64 `validate:"required,gte=0"`
 	// L1BatchTimeoutPerc is the portion of the range before the L1Batch
 	// timeout that will trigger a schedule to forge an L1Batch
-	L1BatchTimeoutPerc float64 `validate:"required"`
+	L1BatchTimeoutPerc float64 `validate:"required,lte=1.0"`
```
ed255, line 76: I suggest adding also `gte=0.0` to force this value to be between 0 and 1

config/config.go
ed255: Actually to follow the previous requirement in the code, this should be `gte=128`

config/config.go
ed255: Ok I've now seen that you added this to follow the previous implementation.   At some point we decoupled the API from the node, and with that change we made these two parameters (`UpdateMetricsInterval` and `UpdateRecommendedFeeInterval`) compulsory, but we forgot to update the code that checks these parameters.  So I think the proper solution would be to just set them both with `validate:"required"`

ed255 (changes requested):

ed255:
Also, please test this change locally by running the node with a configuration that was working previously, to make sure that the validation is not breaking a configuration that was previously working.  (And if a previously working configuration breaks, you should reason why)

Pantani:
@Mikelle, please, don't forget to click in `resolve conversation` buttons for each discussion you already solved, for be easier for us to review again and you track what's was done

Pantani (approved):

config/config.go
```diff
@@ -196,13 +196,13 @@ type Coordinator struct {
 		// GasPriceIncPerc is the percentage increase of gas price set
 		// in an ethereum transaction from the suggested gas price by
 		// the ethereum node
-		GasPriceIncPerc int64
+		GasPriceIncPerc int64 `validate:"gte=0"`
```
Pantani, line 199: if the field is not `required`, the `gte=0` validation does not check, try to set this value to zero and run, the validation goes to pass

Pantani: maybe we can remove it, or use some custom validations

Mikelle: Actually, gte is >= 0, not > 0, so zero should work

Pantani (approved):

ed255 (approved):

Issue #728, opened 2021-04-15
BatchInfo only contains last EthTx


ed255:
## Summary of Bug

BatchInfo contains the field `EthTx` (and `Receipt`) where information about the sent transaction that contains the batch is stored.  The TxManager on certain occasions will resend the batch (sometimes increasing the gas, sometimes increasing the nonce), but the `EthTx` is overwritten!

When incrementing gas/nonce the transaction is different, and thus the TxHash changes.  Because of the overwrite, the previous Tx can't be tracked anymore.  For example, this is problematic in this issue: https://github.com/hermeznetwork/hermez-node/issues/725

### Expected Behavior

All EthTx for sent transactions should be stored and never overwritten.  For example, a batchInfo could have a slice of EthTxs.
When the node sends an ethTx, the corresponding TxHash should be found in a file in the batchesdebug folder if this debug option is enabled (`Debug.BatchPath`).

### Steps to Reproduce

An example of when overwrite happens is when an ethTx is not mined after a timeout, and the TxManager resends the batch with increased gas.

ed255:
Related to https://github.com/hermeznetwork/hermez-node/issues/725

PR #746, opened 2021-04-22
implemented etx txs slice in batchInfo

Mikelle:
Closes https://github.com/hermeznetwork/hermez-node/issues/728

### What does this PR does?

Changed from ethTx to ethTxs in batchInfo.

### DIFF
--- a/coordinator/txmanager.go
+++ b/coordinator/txmanager.go
@@ -258,7 +258,7 @@ func (t *TxManager) sendRollupForgeBatch(ctx context.Context, batchInfo *BatchIn
 	if !resend {
 		t.accNextNonce = auth.Nonce.Uint64() + 1
 	}
-	batchInfo.EthTx = ethTx
+	batchInfo.EthTxs = append(batchInfo.EthTxs, ethTx)
 	log.Infow("TxManager ethClient.RollupForgeBatch", "batch", batchInfo.BatchNum, "tx", ethTx.Hash())
 	now := time.Now()
 	batchInfo.SendTimestamp = now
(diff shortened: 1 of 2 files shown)


coordinator/txmanager.go
```diff
@@ -288,32 +288,38 @@ func (t *TxManager) sendRollupForgeBatch(ctx context.Context, batchInfo *BatchIn
...
-			continue
-		} else if tracerr.Unwrap(err) == ethereum.NotFound {
-			err = nil
-			break
-		} else if err != nil {
-			log.Errorw("TxManager ethClient.EthTransactionReceipt",
-				"attempt", attempt, "err", err)
-		} else {
+	// taking receipt from last transaction
+	for i := len(batchInfo.EthTxs) - 1; i > 0; i-- {
```
Mikelle, line 294: Is it okay, that I take the last transaction to get a receipt?

coordinator/txmanager.go
```diff
@@ -323,16 +329,21 @@ func (t *TxManager) checkEthTransactionReceipt(ctx context.Context, batchInfo *B
 func (t *TxManager) handleReceipt(ctx context.Context, batchInfo *BatchInfo) (*int64, error) {
 	receipt := batchInfo.Receipt
 	if receipt != nil {
-		if batchInfo.EthTx.Nonce()+1 > t.accNonce {
-			t.accNonce = batchInfo.EthTx.Nonce() + 1
+		// taking last eth tx nonce as it probably have the highest nonce
+		lastEthTx := batchInfo.EthTxs[len(batchInfo.EthTxs)-1]
```
Mikelle, line 333: Is it okay, that I take lastEthTx to get nonce?

coordinator/txmanager.go
```diff
@@ -323,16 +329,21 @@ func (t *TxManager) checkEthTransactionReceipt(ctx context.Context, batchInfo *B
...
 		}
 		if receipt.Status == types.ReceiptStatusFailed {
 			batchInfo.Debug.Status = StatusFailed
-			_, err := t.ethClient.EthCall(ctx, batchInfo.EthTx, receipt.BlockNumber)
-			log.Warnw("TxManager receipt status is failed", "tx", receipt.TxHash,
-				"batch", batchInfo.BatchNum, "block", receipt.BlockNumber.Int64(),
-				"err", err)
-			batchInfo.EthTxErr = err
+			var err error
+			for _, tx := range batchInfo.EthTxs {
```
Mikelle, line 340: Is it okay, that I do ethCall for every tx here?

coordinator/txmanager.go
jeffprestes: Yes, because in this task we will not change the way hermeznode manage the nonce. 
So we just need to make sure this change does not break how hermeznode manage the nonces and tx receipts nowadays. 
This change seems to be correct for me but @ed255 's second opinion is very valid too.

coordinator/txmanager.go
jeffprestes: Yes it is, because in this task we will not change the way hermeznode manage the nonce.
So we just need to make sure this change does not break how hermeznode manage the nonces and tx receipts nowadays.
This change seems to be correct for me but @ed255 's second opinion is very valid too.

Pantani (dismissed):

coordinator/txmanager.go
OBrezhniev: Ok

OBrezhniev (changes requested):
Looks good to me, just a small fix needed.

ARR552 (approved):

jeffprestes (approved):
It seems to be OK.
However, this would require important tests and monitoring the coordinador's behavior for couple of days with several tx in internaltestnet

Pantani (approved):

OBrezhniev (approved):

Issue #739, opened 2021-04-20
Avoid to create unnecessary accounts


jeffprestes:
## Summary of Bug

Duplicating accounts with the same BJJ address with same tokenID

### Expected Behavior

When /transaction-pool receives a new tx where the recipient (to field) is a BJJ address, it should perform a query in account table in database for this BJJ address before to return an error and asking to the user to register it at /account-creation-authorization

In account table it's possible to an account to have a BJJ and ETH address set and the tx sender could be allowed to send L2 tx to any account using either BJJ or ETH address without creates a new account.

### Steps to Reproduce

To check this please verify the hez:0x7CaEFD0214dCd8305A7991eB8E016d3a76822189 in Goerli/Internaltestnet

Pantani:
`TransferToBjj` don't have this verification, only for `TransferToEth` and `Transfer`

https://docs.hermez.io/#/developers/protocol/hermez-protocol/protocol?id=account-types

Pantani:
If we merge this implementation, we can break some exchanges integrations

jeffprestes:
After meeting with Jordi, Carlos and Bennassar we checked the user can send tokens to a BJJ that is not already created but internally the before the hermez-node create a new account (and/or leaf in Merkle Tree) it needs to check if the account already exists and reuse it. The API must accept the transaction and handle this internally.

arnaubennassar:
Closed in favor of #680

PR #755, opened 2021-04-27
added checker for toBjj address in transaction creation

Mikelle:
Closes https://github.com/hermeznetwork/hermez-node/issues/739

### What does this PR does?

added checker for toBjj address in transaction creation

### DIFF
--- a/db/l2db/apiqueries.go
+++ b/db/l2db/apiqueries.go
@@ -69,6 +69,16 @@ func (l2db *L2DB) AddTxAPI(tx *PoolL2TxWrite) error {
 			feeUSD, l2db.maxFeeUSD))
 	}
 
+	if tx.ToBJJ != nil {
+		var exists bool
+		row := l2db.dbRead.QueryRow(`SELECT EXISTS(SELECT 1 FROM account WHERE bjj = $1);`, tx.ToBJJ)
+		if err := row.Scan(&exists); err != nil {
+			return tracerr.Wrap(err)
+		} else if !exists {
+			return tracerr.Wrap(fmt.Errorf("account with bjj %s doesn't exists. "+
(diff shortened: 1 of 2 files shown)


db/l2db/apiqueries.go
```diff
@@ -69,6 +69,16 @@ func (l2db *L2DB) AddTxAPI(tx *PoolL2TxWrite) error {
 			feeUSD, l2db.maxFeeUSD))
 	}
 
+	if tx.ToBJJ != nil {
+		var exists bool
+		row := l2db.dbRead.QueryRow(`SELECT EXISTS(SELECT 1 FROM account WHERE bjj = $1);`, tx.ToBJJ)
+		if err := row.Scan(&exists); err != nil {
+			return tracerr.Wrap(err)
+		} else if !exists {
```
jeffprestes, line 77: Maybe split this if/else chain in two verifications would it make the code more clean. 
One if verifies the error
Other if, after the value is set to exists variable, check the exists variable content.

Pantani (changes requested):
Let's wait for the answer to it:
https://github.com/hermeznetwork/hermez-node/issues/739#issuecomment-828370528

jeffprestes:
After meeting with Jordi, Carlos and Bennassar we checked the user can send tokens to a BJJ that is not already created but internally the before the hermez-node create a new account it needs to check if the account already exists and reuse it. So, this implementations needs to be made in tx-selector or batch builder and not in API. The API must accept the tx.

Pantani:
@jeffprestes @Mikelle, can we close this one?

Mikelle:
@Pantani I think yes, this pr is no longer relevant, bcs API have to accept such transactions

PR #761, opened 2021-04-29
Fix freezes in the coordinator mode in case of the prover connection issue

### LINKED ISSUE #745: Synchronizer becomes stuck when starting the pipeline and the proof server is forever "busy"
### Unbuffered synchronizer -> coordinator channel
The channel used to send synchronization events to the coordinator is not buffered.  This means that while the coordinator is busy and not waiting at the channel, the synchronizer blocks:
https://github.com/hermeznetwork/hermez-node/blob/616249dcf508c04c10ddd6228792869ab004134a/coordinator/coordinator.go#L289
The `msgCh` channel should be buffered, with a proper value for the buffer size, so that if some times the coordinator takes longer, the synchronizer can continue synchronizing.

### No timeout to wait for proof server to be ready
When creating the pipeline (which is done by the main coordinator routine), there's no timeout to wait for the proof servers to be ready: https://github.com/hermeznetwork/hermez-node/blob/f4f77aa84bd8a55fdc057fde6de9d60fac9173b7/coordinator/pipeline.go#L99
If the proof server stays in the "busy" state forever (which is a bug of the proof server), the hermez-node will become stuck forever.  It would be better to have a timeout.  Since `prover.WaitReady` takes a context, I think the best solution is to wrap this context with a timeout for each proof server, so that the waiting time has a timeout.  This timeout could be set by configuration.

stunndard:
Closes #745

### What does this PR does?

Fix freezes with busy/broken prover.
Remove blocking of synchronizer by coordinator.

### DIFF
--- a/coordinator/coordinator.go
+++ b/coordinator/coordinator.go
@@ -276,7 +276,7 @@ func NewCoordinator(cfg Config,
 
 		purger: &purger,
 
-		msgCh: make(chan interface{}),
+		msgCh: make(chan interface{}, queueLen),
 		ctx:   ctx,
 		// wg
 		cancel: cancel,
--- a/coordinator/pipeline.go
+++ b/coordinator/pipeline.go
@@ -22,6 +22,9 @@ import (
 	"github.com/hermeznetwork/tracerr"
 )
 
+// TODO: add this timeout to config file?
+const proverWaitReadyTimeout = 20 * time.Second
+
 type statsVars struct {
 	Stats synchronizer.Stats
 	Vars  common.SCVariablesPtr
@@ -97,7 +100,9 @@ func NewPipeline(ctx context.Context,
 	proversPool := NewProversPool(len(provers))
 	proversPoolSize := 0
 	for _, prover := range provers {
-		if err := prover.WaitReady(ctx); err != nil {
+		ctxTimeout, ctxTimeoutCancel := context.WithTimeout(ctx, proverWaitReadyTimeout)
+		defer ctxTimeoutCancel()
+		if err := prover.WaitReady(ctxTimeout); err != nil {
 			log.Errorw("prover.WaitReady", "err", err)
 		} else {
 			proversPool.Add(ctx, prover)


coordinator/coordinator.go
```diff
@@ -276,7 +276,7 @@ func NewCoordinator(cfg Config,
 
 		purger: &purger,
 
-		msgCh: make(chan interface{}),
+		msgCh: make(chan interface{}, 16),
```
jeffprestes, line 279: Maybe leave the size of the buffer in a separated variable and use it in the channel creation. Also, I'd leave a TODO comment on top of it to further analysis to check if the size of the buffer must reside or not in config file.

coordinator/pipeline.go
```diff
@@ -97,7 +97,9 @@ func NewPipeline(ctx context.Context,
 	proversPool := NewProversPool(len(provers))
 	proversPoolSize := 0
 	for _, prover := range provers {
-		if err := prover.WaitReady(ctx); err != nil {
+		ctxTimeout, cancel := context.WithTimeout(ctx, 20*time.Second)
```
jeffprestes, line 100: Maybe leave the interval of the timeout in a separated variable and use it in the context creation. Also, I'd leave a TODO comment on top of it to further analysis to check whether the size of the interval must reside or not in config file.

stunndard: Uploaded a new version

jeffprestes (approved):

coordinator/pipeline.go
```diff
@@ -22,6 +22,9 @@ import (
 	"github.com/hermeznetwork/tracerr"
 )
 
+// TODO: add this timeout to config file?
```
arnaubennassar, line 25: Suggestion: remove this coment and open an issue to add this in the config later on.

Reasoning: I'm not sure if we've implemented some sort of default values vars that come from the config file. If not adding this would need to be required which would break the already deployed nodes using the config file as it is. If this is the case, and we follow `semver` we would be forced to do a major update.

If you agree on the suggestion, please in the issue mention to investigate the possibility of default values.

arnaubennassar (approved):
Please, consider the suggestion before merging

PR #747, opened 2021-04-22
Feature/fastsync concurrent block fetching

OBrezhniev:
Closes #740

### What does this PR does?

I've done decoupling of fetching and processing as a first step, then made fetching blocks and events in large batches (batchSize=100, but limited to max 10 concurrent block header requests) inside the sync loop if there's more then 256+batchSize blocks to synchronize, fetch events in large batches, fetch blocks in concurrent requests and process them after.

### How to test?

Run the node and see how it synchronizes.

## Todo

- [ ] Add config options for magic numbers
- [ ] Update unit tests to cover new functions and changed work flows of existing

PR #766, opened 2021-04-30
Feature/fastsync make checkpoint

OBrezhniev:
### What does this PR does?

Changes how checkpointing is done in synchronization. Based on #747, when block range is synchronized calls MakeCheckpoint only once for the whole range and only if there were batches.

Changes storage consistency assumptions. Previously checkpoints were always done before writing to PostgreSQL, now - after in case of fast sync of block range.

### How to test?

Test performance by running node synchronization.
Test that node correctly shuts down on Ctrl+C and continues synchronization after that.

## Checklist

TODO:
- [ ] Check how node behaves on forced termination of it's process.


jeffprestes:
Closing it as we're going to review sync process for zkEVM.

PR #773, opened 2021-04-30
Feature/#729 gasprice etherscan

### LINKED ISSUE #729: Use EtherScan gas estimations
## Summary
Use EtherScan gas estimations rather than ethereum node estimations because etherScan provide more accurate estimations than an ethereum node that can  provide outdated estimations under some circunstances

### Expected Behavior
Gas estimations should be more accurate when gas is growing quickly to prevent ethereum tx (containing batches) from failing

ARR552:
Closes #729

### What does this PR does?

This PR adds the ability to retrieve the gas price estimation from etherscan which is more accurate. If the endpoint is not operative then the node uses the old method as backup which is ask the gas price to geth.

There are four different situations:
1. EtherScan variables are empty or missing in config file: In this case the node will work only asking the gas price to geth.
2. EtherScan variables are wrong. In this case the node will try to use them. As the api call will fail, the node will use the geth method to retrieve the gas price.
3. EtherScan variables are correct but the endpoint is no available. In this situation the node will try to get the gas price from etherscan. Due to the endpoint unavailability, this call will fail and the node will try the geth method.
4. If Etherscan variables are correct and the endpoint is working, the node will retrieve the gas price info from etherscan.

This PR adds new configuration parameters:
- Coordinator.Etherscan.URL
- Coordinator.Etherscan.APIKey

### How to test?
Run the hermez node in coordinator mode and let the node ask for the gas price before send the ethereum transaction.

## Checklist

review them:

- [X] Include tests

### DIFF
--- a/coordinator/txmanager.go
+++ b/coordinator/txmanager.go
@@ -25,15 +25,15 @@ import (
 // call to forge, waits for transaction confirmation, and keeps checking them
 // until a number of confirmed blocks have passed.
 type TxManager struct {
-	cfg             Config
-	ethClient       eth.ClientInterface
-	etherscanClient etherscan.Client
-	l2DB            *l2db.L2DB   // Used only to mark forged txs as forged in the L2DB
-	coord           *Coordinator // Used only to send messages to stop the pipeline
-	batchCh         chan *BatchInfo
-	chainID         *big.Int
-	account         accounts.Account
-	consts          common.SCConsts
+	cfg              Config
+	ethClient        eth.ClientInterface
+	etherscanService *etherscan.Service
+	l2DB             *l2db.L2DB   // Used only to mark forged txs as forged in the L2DB
+	coord            *Coordinator // Used only to send messages to stop the pipeline
+	batchCh          chan *BatchInfo
+	chainID          *big.Int
+	account          accounts.Account
+	consts           common.SCConsts
 
 	stats       synchronizer.Stats
 	vars        common.SCVariables
@@ -57,7 +57,7 @@ type TxManager struct {
 
 // NewTxManager creates a new TxManager
 func NewTxManager(ctx context.Context, cfg *Config, ethClient eth.ClientInterface, l2DB *l2db.L2DB,
-	coord *Coordinator, scConsts *common.SCConsts, initSCVars *common.SCVariables, etherscanClient etherscan.Client) (
+	coord *Coordinator, scConsts *common.SCConsts, initSCVars *common.SCVariables, etherscanService *etherscan.Service) (
 	*TxManager, error) {
 	chainID, err := ethClient.EthChainID()
 	if err != nil {
@@ -75,7 +75,7 @@ func NewTxManager(ctx context.Context, cfg *Config, ethClient eth.ClientInterfac
 	return &TxManager{
 		cfg:               *cfg,
 		ethClient:         ethClient,
-		etherscanClient:   etherscanClient,
+		etherscanService:  etherscanService,
 		l2DB:              l2DB,
 		coord:             coord,
 		batchCh:           make(chan *BatchInfo, queueLen),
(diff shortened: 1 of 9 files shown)


jeffprestes (commented):
As per our conversation @ARR552 , it would be great perform small changes in the code to make this new feature totally optional to the coordinator just leaving either the Etherscan.URL or Etherscan.APIKEY fields blank at cfg.buidler.toml .

jeffprestes (dismissed):

coordinator/txmanager.go
```diff
@@ -126,9 +129,37 @@ func (t *TxManager) syncSCVars(vars common.SCVariablesPtr) {
...
-	gasPrice, err := t.ethClient.EthSuggestGasPrice(ctx)
-	if err != nil {
-		return nil, tracerr.Wrap(err)
+	//First we try getting the gas price from etherscan. If it fails we get the gas price from the ethereum node.
+	var gasPrice *big.Int
+	if t.etherscanService != nil {
+		etherscanGasPrice, err := (*t.etherscanService).GetGasPrice(ctx)
+		if err != nil {
+			log.Warn("Error getting the gas price from etherscan. Trying another method. Error: ", err.Error())
+			var er error
```
Mikelle, line 138: Why do you need to create an additional variable **er**, and not use the previously declared variable **err**?

coordinator/txmanager.go
```diff
@@ -126,9 +129,37 @@ func (t *TxManager) syncSCVars(vars common.SCVariablesPtr) {
...
+				return nil, tracerr.Wrap(er)
+			}
+		} else {
+			eGasPrice := new(big.Int)
+			eGasPrice, ok := eGasPrice.SetString(etherscanGasPrice.ProposeGasPrice, 10)
+			if !ok {
+				log.Warn("invalid big int: \"%v\"", etherscanGasPrice.ProposeGasPrice, ". Trying another method to get gas price")
+				var er error
+				gasPrice, er = t.ethClient.EthSuggestGasPrice(ctx)
+				if er != nil {
```
Mikelle, line 150: Why do you need to create an additional variable **er**, and not use the previously declared variable **err**?

coordinator/txmanager.go
```diff
@@ -126,9 +129,37 @@ func (t *TxManager) syncSCVars(vars common.SCVariablesPtr) {
...
+					return nil, tracerr.Wrap(er)
+				}
+			} else {
+				gasPrice = eGasPrice
+			}
+		}
+	} else {
+		var er error
```
Mikelle, line 158: Why do you need to create an additional variable **er**, and not use the previously declared variable **err**?

node/node.go
```diff
@@ -325,6 +326,15 @@ func NewNode(mode Mode, cfg *config.Node, version string) (*Node, error) {
...
 			return nil, tracerr.Wrap(err)
 		}
+		var etherScanService *etherscan.Service
+		if cfg.Coordinator.Etherscan.URL != "" && cfg.Coordinator.Etherscan.APIKey != "" {
+			log.Info("EtherScan method detected in cofiguration file")
+			etherScanService, _ = etherscan.NewEtherscanService(cfg.Coordinator.Etherscan.URL,
+				cfg.Coordinator.Etherscan.APIKey)
+		} else {
+			log.Info("EtherScan method not configured in config file")
+			etherScanService = nil
```
Mikelle, line 336: Seems that putting nil there is redundant, bcs etherScanService already nil.

coordinator/txmanager.go
```diff
@@ -126,9 +129,37 @@ func (t *TxManager) syncSCVars(vars common.SCVariablesPtr) {
 
 // NewAuth generates a new auth object for an ethereum transaction
 func (t *TxManager) NewAuth(ctx context.Context, batchInfo *BatchInfo) (*bind.TransactOpts, error) {
-	gasPrice, err := t.ethClient.EthSuggestGasPrice(ctx)
-	if err != nil {
-		return nil, tracerr.Wrap(err)
+	//First we try getting the gas price from etherscan. If it fails we get the gas price from the ethereum node.
+	var gasPrice *big.Int
+	if t.etherscanService != nil {
+		etherscanGasPrice, err := (*t.etherscanService).GetGasPrice(ctx)
```
Pantani, line 135: Why is this casting needed?

```suggestion
		etherscanGasPrice, err := t.etherscanService.GetGasPrice(ctx)
```

coordinator/txmanager.go
```diff
@@ -126,9 +129,37 @@ func (t *TxManager) syncSCVars(vars common.SCVariablesPtr) {
...
+			var er error
+			gasPrice, er = t.ethClient.EthSuggestGasPrice(ctx)
+			if er != nil {
+				return nil, tracerr.Wrap(er)
+			}
+		} else {
+			eGasPrice := new(big.Int)
+			eGasPrice, ok := eGasPrice.SetString(etherscanGasPrice.ProposeGasPrice, 10)
+			if !ok {
+				log.Warn("invalid big int: \"%v\"", etherscanGasPrice.ProposeGasPrice, ". Trying another method to get gas price")
```
Pantani, line 147: Try to avoid long lines. You can break for a better clean code

coordinator/txmanager.go
```diff
@@ -126,9 +129,37 @@ func (t *TxManager) syncSCVars(vars common.SCVariablesPtr) {
...
 func (t *TxManager) NewAuth(ctx context.Context, batchInfo *BatchInfo) (*bind.TransactOpts, error) {
-	gasPrice, err := t.ethClient.EthSuggestGasPrice(ctx)
-	if err != nil {
-		return nil, tracerr.Wrap(err)
+	//First we try getting the gas price from etherscan. If it fails we get the gas price from the ethereum node.
+	var gasPrice *big.Int
+	if t.etherscanService != nil {
+		etherscanGasPrice, err := (*t.etherscanService).GetGasPrice(ctx)
+		if err != nil {
+			log.Warn("Error getting the gas price from etherscan. Trying another method. Error: ", err.Error())
```
Pantani, line 137: Try to avoid long lines. You can break for a better clean code

node/node.go
ARR552: Yes, I know. It just to make it more readable

coordinator/txmanager.go
```diff
@@ -126,9 +129,37 @@ func (t *TxManager) syncSCVars(vars common.SCVariablesPtr) {
 
 // NewAuth generates a new auth object for an ethereum transaction
 func (t *TxManager) NewAuth(ctx context.Context, batchInfo *BatchInfo) (*bind.TransactOpts, error) {
-	gasPrice, err := t.ethClient.EthSuggestGasPrice(ctx)
-	if err != nil {
-		return nil, tracerr.Wrap(err)
+	//First we try getting the gas price from etherscan. If it fails we get the gas price from the ethereum node.
+	var gasPrice *big.Int
+	if t.etherscanService != nil {
```
Pantani, line 134: _maybe it's better creates a logic only to get the gas price, and we can create unit tests to handle this logic_

etherscan/etherscan_test.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"context"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+var etherscanURL = "http://localhost:3000/api"
```
Pantani, line 12: Why `localhost`? Is there some way to mock it locally?

node/node.go
```diff
@@ -325,6 +326,15 @@ func NewNode(mode Mode, cfg *config.Node, version string) (*Node, error) {
 		if err != nil {
 			return nil, tracerr.Wrap(err)
 		}
+		var etherScanService *etherscan.Service
+		if cfg.Coordinator.Etherscan.URL != "" && cfg.Coordinator.Etherscan.APIKey != "" {
+			log.Info("EtherScan method detected in cofiguration file")
+			etherScanService, _ = etherscan.NewEtherscanService(cfg.Coordinator.Etherscan.URL,
```
Pantani, line 332: Question: why not handle this error?

coordinator/txmanager.go
jeffprestes: I agree. 
However, could we set to 160 columns?
80 columns were for old punch cards https://softwareengineering.stackexchange.com/questions/262723/why-were-punch-cards-used-for-programming  or for 80' or 90' ancient monitors.

jeffprestes (approved):

PR #791, opened 2021-05-05
Make txselector ignore exits with amount 0

arnaubennassar:
**EDIT:** Repurposed the PR to filter this txs a txselector level

### What does this PR does?

Make txselector ignore transactions of type exit with amount 0

## Checklist

review them:

- [ ] Include tests

### DIFF
--- a/txprocessor/txprocessor.go
+++ b/txprocessor/txprocessor.go
@@ -418,7 +418,7 @@ func (tp *TxProcessor) ProcessTxs(coordIdxs []common.Idx, l1usertxs, l1coordinat
 			}
 		}
 		if tp.s.Type() == statedb.TypeSynchronizer || tp.s.Type() == statedb.TypeBatchBuilder {
-			if exitIdx != nil && exitTree != nil {
+			if exitIdx != nil && exitTree != nil && exitAccount != nil {
 				exits[tp.i] = processedExit{
 					exit:    true,
 					newExit: newExit,
--- a/txselector/txselector.go
+++ b/txselector/txselector.go
@@ -358,6 +358,13 @@ func (txsel *TxSelector) processL2Txs(tp *txprocessor.TxProcessor,
 			break
 		}
 
+		// Discard exits with amount 0
+		if l2Txs[i].Type == common.TxTypeExit && l2Txs[i].Amount.Cmp(big.NewInt(0)) <= 0 {
+			l2Txs[i].Info = "Exits with amount 0 have no sense, not accepting to prevent unintended transactions"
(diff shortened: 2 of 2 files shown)


txprocessor/txprocessor.go
```diff
@@ -395,6 +395,10 @@ func (tp *TxProcessor) ProcessTxs(coordIdxs []common.Idx, l1usertxs, l1coordinat
 
 	// Process L2Txs
 	for i := 0; i < len(l2txs); i++ {
+		if l2txs[i].Type == common.TxTypeExit && l2txs[i].Amount.Cmp(big.NewInt(0)) == 0 {
```
jeffprestes, line 398: It seems to be OK.
However, is there a way to give any feedback to the user? Maybe block this in API level too?

jeffprestes (approved):

txprocessor/txprocessor.go
Pantani, line 398: Is there any way to the amount be negative? Maybe we should filter this way:

```suggestion
		if l2txs[i].Type == common.TxTypeExit && l2txs[i].Amount.Cmp(big.NewInt(0)) <= 0 {
```

Pantani: Also, can we test this case?

krlosMata:
Could you please double check on this scenario:
 - `txprocessor.go` is used in `sync` mode
 - assume that an alternative node process a L2 exit transaction with amount = 0
 -  Proper data-availability would be on the smart contract. In this case `From = X | To = 1(exit_idx) | Amount = 0  | Fee = X`
 - `hermez-node` should process this L2 exit transaction with amount = 0 to build the new state tree, since the sender account will increase its nonce by 1

jeffprestes:
After @krlosMata comments, could you please , @arnaubennassar , test this scenario in sync mode?

jeffprestes (changes requested):
Test sync mode with another node processing exit with amount 0 seems to be appropriated

txprocessor/txprocessor.go
arnaubennassar: Marking as resolved as you already approved the PR that will do just do it

jeffprestes (commented):
It seems to be OK. However, I strongly recommend to test this change in internaltestnet before to merge.

Pantani (approved):

arnaubennassar:
> It seems to be OK. However, I strongly recommend to test this change in internaltestnet before to merge.

Tested successfully on internaltestnet

jeffprestes (approved):

PR #795, opened 2021-05-07
BatchNum added to txpool info field

### LINKED ISSUE #762: Add batch num in info messages for pool txs
## Rationale

Some times a tx has a `info` field seted that may be outdated. It would be nice to add to the message which batch was the coordinator forging when modified the info value.

## Implementation

Just modify the string to include the batch num

ARR552:
Closes #762

### What does this PR does?

This PR adds the batch number to the info field in tx_pool database table to give us the ability to check when the txselector updates the info message.

### How to test?

Run the node in coordinator mode and send an invalid tx.

## Checklist

- [ ] Include tests


arnaubennassar:
Suggestion: add `batchNum` argument in `func (l2db *L2DB) UpdateTxsInfo(txs []common.PoolL2Tx) error` ad edit the function so it adds the batchNum at the beginning of every message. 
This way you only write once the batchNum stuff instead of for evcery different message. Furthermore it's a nice way to ensure consistency for all the messages, and it's easier to modify in the future.

db/l2db/l2db.go
```diff
@@ -125,7 +126,7 @@ func (l2db *L2DB) UpdateTxsInfo(txs []common.PoolL2Tx) error {
 	}
 	txUpdates := make([]txUpdate, len(txs))
 	for i := range txs {
-		txUpdates[i] = txUpdate{ID: txs[i].TxID, Info: txs[i].Info}
+		txUpdates[i] = txUpdate{ID: txs[i].TxID, Info: "BatchNum: " + strconv.FormatInt(int64(batchNum), 10) + ". " + txs[i].Info}
```
arnaubennassar, line 129: since `"BatchNum: " + strconv.FormatInt(int64(batchNum), 10) + ". "` will be the same for all the iterations of the loop, it would be better to save this in a var outside the loop

arnaubennassar (changes requested):

Mikelle (approved):

arnaubennassar (approved):

PR #796, opened 2021-05-07
Add CLI command to downgrade SQL schema

### LINKED ISSUE #741: Add tests for SQL schema migrations
## Rationale

Right now changes on the SQL schema don't have proper unit tests. A wrong migration could force the node to re-synchronize from scratch, lose data (from `tx_pool` and `account_creation_auth`) and even completely break the node.

Keep in mind that unit tests will work with the latest schema well as each of them wipes the DB and adds fresh new data before starting. But this means that the scenario where the DB has data and goes through a migration is not tested.

## Implementation

Add a new unit test file (probably under `db/migrations`), this test should:
1. Run the first migration
2. Insert data in the tables (we could do that only for tables that got affected by subsequent migrations, and keep adding when needed)
3. Run the rest of migrations
4. Check that the data added in step 2 is correct (still there with the correct content)

### LINKED ISSUE #794: Add option to downgrade DB schema version
## Rationale

The binary takes care of updating DB schema however there is not a mechanism to revert those changes, which could be needed if there is some unexpected behavior when updating versions of the node

## Implementation

Add a cli command to downgrade to a specific migration

arnaubennassar:
Closes #794 
Closes #741 

### What does this PR does?

- Add a CLI option to revert SQL migrations (migrate down)
- Add test utilities for testing SQL migrations
- Add test for the existing migration (`db/migrations/0002.sql`)

### How to test?

- Migrations are tested with unit test
- CLI option can be tested as follows:

1. Deploy a node in coordinator mode (using this branch)
2. Send at least one tx to the pool
3. Stop the node
4. Run the command `heznode migratesqldown -mode=sync -cfg=<your config file> -nMigrations=1`
5. Check that the DB has no `item_id` column and the txs you sent are still there
6. Run again the node, this time sing a version that didn't had the 2nd migration file (for instance `v1.0.0`)
7. Check that the node behaves well

## Checklist

review them:

- [x] Include tests


cmd/heznode/main.go
```diff
@@ -212,6 +214,48 @@ func cmdWipeDBs(c *cli.Context) error {
...
+		cfg.PostgreSQL.NameWrite,
+	)
+	if err != nil {
+		return tracerr.Wrap(err)
+	}
+	log.Infof("Reverting %d SQL migrations...", migrationsToRun)
+	if err := dbUtils.MigrationsDown(db.DB, migrationsToRun); err != nil {
+		return tracerr.Wrap(fmt.Errorf("dbUtils.MigrationsDown: %w", err))
+	}
+	log.Info("SQL migrations run successfully")
```
Pantani, line 254:
```suggestion
	log.Info("SQL migrations down successfully")
```

ARR552 (dismissed):

cmd/heznode/main.go
arnaubennassar: Applied this change

PR #817, opened 2021-05-14
Add test for 3rd migration

### LINKED ISSUE #815: Add test for 0003.sql migration
## Rationale

This migration is not tested

arnaubennassar:
Closes #815 

### What does this PR does?

Add unit test for the 3rd migration (`db/migrations/0003.sql`)

### How to test?

Just run the unit tests

## Checklist

review them:

- [x] Include tests

### DIFF
--- a/db/migrations/0003_test.go
+++ b/db/migrations/0003_test.go
@@ -1,0 +1,104 @@
+package migrations_test
+
+import (
+	"testing"
+
+	"github.com/jmoiron/sqlx"
+	"github.com/stretchr/testify/assert"
+)
+
+// This migration adds the column `eth_tx_hash` on `batch`
+
+type migrationTest0003 struct{}
+
+func (m migrationTest0003) InsertData(db *sqlx.DB) error {
+	// insert block to respect the FKey of batch
+	const queryInsertBlock = `INSERT INTO block (
+		eth_block_num,"timestamp",hash
+	) VALUES (
+		4417296,'2021-03-10 16:44:06.000',decode('C4D46677F3B2511D96389521C2BDFFE91127DE214423FF14899A6177631D2105','hex')
+	);`
+	// insert batch
+	const queryInsertBatch = `INSERT INTO batch (
+		batch_num, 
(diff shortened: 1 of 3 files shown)


db/migrations/0003_test.go
```diff
@@ -0,0 +1,104 @@
...
+		717, 
+		115.047487133272
+	);`
+	_, err := db.Exec(queryInsertBlock + queryInsertBatch)
+	return err
+}
+
+func (m migrationTest0003) RunAssertsAfterMigrationUp(t *testing.T, db *sqlx.DB) {
+	// check that the batch inserted in previous step is persisted with same content
+	const queryGetBatch = `SELECT COUNT(*) FROM batch WHERE
```
Mikelle, line 55: maybe use `select exists(select 1 from batch...)`? Count(*) is not really efficient.

db/migrations/0003_test.go
```diff
@@ -0,0 +1,104 @@
...
+	`
+	row := db.QueryRow(queryGetBatch)
+	var result int
+	assert.NoError(t, row.Scan(&result))
+	assert.Equal(t, 1, result)
+}
+
+func (m migrationTest0003) RunAssertsAfterMigrationDown(t *testing.T, db *sqlx.DB) {
+	// check that the batch inserted in previous step is persisted with same content
+	const queryGetBatch = `SELECT COUNT(*) FROM batch WHERE
```
Mikelle, line 78: maybe use select exists(select 1 from batch...)? Count(*) is not really efficient.

Mikelle (changes requested):

db/migrations/0003_test.go
arnaubennassar: Since this is a test that will always have a single row, performance is not an issue

db/migrations/0003_test.go
arnaubennassar: Since this is a test that will always have a single row, performance is not an issue

Mikelle (approved):

PR #823, opened 2021-05-19
Implement atomic-pool endpoint

### LINKED ISSUE #769: [AtomicTx] API endpoint
## Rationale

We must create a new API for the node to receive these transactions and validate them. The API must validate, also, the group of transactions together. One of these transactions can spend from another atomic transaction, from an account without balance before this transaction. The user must batch all signed transactions in the same request to broadcast all of them to be atomic.

## Implementation

We will create a new API method for sending atomic transactions called `atomic-pool`. This API must receive an array of transactions and validate all, also. If the sum of the fees is greater than the minimum node fee after, it will add the transaction into the `tx_pool` database, linking each other by the TxID. We must update the swagger files as well.

e.g.:

```bash
$ curl -X POST "<http://localhost:8086/v1/atomic-pool>" \
-H  "Content-Type: application/json" \
-d '[
  {
    "id": "0x029761c516998ab5261ca388574e5442ff926ce19ff2497e646f6bd58d66669b85",
    "type": "Transfer",
    "tokenId": 0,
    "fromAccountIndex": "hez:ETH:1276",
    "toAccountIndex": "hez:ETH:256",
    "toHezEthereumAddress": "hez:0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF",
    "toBjj": "hez:AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
    "amount": "10",
    "fee": 126,
    "nonce": 53,
    "signature": "494658979f0b4c93081a8aa6fd798c9b895c72b60cd3882a46339c564319a22a633d7a148b0cd25da537879478876204486dc38a3dcf0ce1b5717cbbbac56a02"
  },
  {
    "id": "0x0252222e0de00610eb3d32a3cbe72020ab8810c3518140e3bf82f1be663b26f1d2",
    "type": "Transfer",
    "tokenId": 1,
    "fromAccountIndex": "hez:ETH:256",
    "toAccountIndex": "hez:ETH:1276",
    "toHezEthereumAddress": "hez:0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF",
    "toBjj": "hez:AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
    "amount": "5",
    "fee": 126,
    "nonce": 30,
    "signature": "494658979f0b4c93081a8aa6fd798c9b895c72b60cd3882a46339c564319a22a633d7a148b0cd25da537879478876204486dc38a3dcf0ce1b5717cbbbac56a02"
  }
]'
```

### Assumptions

- All transaction must be signed by the destination address from itself;
- The endpoint `/atomic-pool` do not accept single transactions;
- All transactions sent into the endpoint `/atomic-pool` must have the same transaction validations into the pool, besides the fee;
- The total fee must be a sum of all transactions fee;
- The user can be the owner of all transactions and send all of them already signed;
- User A can make a simple transfer to user B and pay the fee using another token;
- `rqTxID` must be a valid `TxID`;
- The `rq` fields will be filled by the node;
- The transactions do not know about others in the protocol level, only in the node was broadcasted;

### TODOs

- [ ] Create new API method;
- [ ] Add transaction validations;
- [ ] Insert into the database;
- [ ] Calculate the fee from all tx's together;
- [ ] Update swagger files;

arnaubennassar:
Closes #769 

### What does this PR does?

Implement `POST /atomic-pool` endpoint for clients to be able to send atomic groups of transactions (all get forged or non get forged)

### How to test?

- Unit tests included
- Spin up the server and use the endpoint

## Checklist

review them:

- [x] Include tests
- [x] Update swagger file (if needed)

### DIFF
--- a/api/txspool.go
+++ b/api/txspool.go
@@ -13,6 +13,7 @@ import (
 	"github.com/hermeznetwork/hermez-node/db/l2db"
 	"github.com/hermeznetwork/tracerr"
 	"github.com/iden3/go-iden3-crypto/babyjub"
+	"github.com/yourbasic/graph"
 )
 
 func (a *API) postPoolTx(c *gin.Context) {
@@ -30,12 +31,86 @@ func (a *API) postPoolTx(c *gin.Context) {
 	}
 	writeTx.ClientIP = c.ClientIP()
 	// Insert to DB
+	// note that fee in USD validation happens in this function
 	if err := a.l2.AddTxAPI(writeTx); err != nil {
 		retSQLErr(err, c)
 		return
 	}
 	// Return TxID
 	c.JSON(http.StatusOK, writeTx.TxID.String())
+}
+
+func (a *API) postAtomicPool(c *gin.Context) {
+	// Parse body
+	var receivedTxs []receivedPoolTx
+	if err := c.ShouldBindJSON(&receivedTxs); err != nil {
+		retBadReq(err, c)
+		return
+	}
+	nTxs := len(receivedTxs)
+	if nTxs <= 1 {
+		retBadReq(errors.New(ErrSingleTxInAtomicEndpoint), c)
+		return
+	}
+	// Transform from received to insert format and validate (individually)
+	writeTxs := make([]l2db.PoolL2TxWrite, nTxs) // used for DB insert
+	txIDStrings := make([]string, nTxs)          // used for successful response
+	clientIP := c.ClientIP()
+	for i, tx := range receivedTxs {
+		writeTx := tx.toPoolL2TxWrite()
+		if err := a.verifyPoolL2TxWrite(writeTx); err != nil {
+			retBadReq(err, c)
+			return
+		}
+		writeTx.ClientIP = clientIP
+		writeTxs[i] = *writeTx
+		txIDStrings[i] = writeTx.TxID.String()
+	}
+	// Validate that all txs in the payload represent an atomic group
+	if !isAtomicGroup(writeTxs) {
+		retBadReq(errors.New(ErrTxsNotAtomic), c)
+		return
+	}
+	// Insert to DB
+	if err := a.l2.AddAtomicTxsAPI(writeTxs); err != nil {
+		retSQLErr(err, c)
+		return
+	}
+	// Return IDs of the added txs in the pool
+	c.JSON(http.StatusOK, txIDStrings)
+}
+
+// isAtomicGroup returns true if all the txs are needed to be forged
+// (all txs will be forged in the same batch or non of them will be forged)
+func isAtomicGroup(txs []l2db.PoolL2TxWrite) bool {
+	// Create a graph from the given txs to represent requests between transactions
+	g := graph.New(len(txs))
+	idToPos := make(map[common.TxID]int, len(txs))
+	// Map tx ID to integers that will represent the nodes of the graph
+	for i, tx := range txs {
+		idToPos[tx.TxID] = i
+	}
+	// Create vertices that connect nodes of the graph (txs) using RqTxID
+	for i, tx := range txs {
+		if tx.RqTxID == nil {
+			// if just one tx doesn't request any other tx, this tx could be forged alone
+			// making the hole group not atomic
+			return false
+		}
+		if rqTxPos, ok := idToPos[*tx.RqTxID]; ok {
+			g.Add(i, rqTxPos)
+		} else {
+			// tx is requesting a tx that is not provided in the payload
+			return false
+		}
+	}
+	// A graph with a single strongly connected component,
+	// means that all the nodes can be reached from all the nodes.
+	// If tx A "can reach" tx B it means that tx A requests tx B.
+	// Therefore we can say that if there is a single strongly connected component in the graph,
+	// all the transactions require all trnsactions to be forged, in other words: they are an atomic group
+	strongComponents := graph.StrongComponents(g)
+	return len(strongComponents) == 1
 }
 
 func (a *API) getPoolTx(c *gin.Context) {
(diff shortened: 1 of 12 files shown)


db/l2db/apiqueries.go
```diff
@@ -104,11 +105,123 @@ func (l2db *L2DB) AddTxAPI(tx *PoolL2TxWrite) error {
...
+		TokenID     common.TokenID `meddler:"token_id"`
+		USDPerToken float64        `meddler:"usd_no_decimals"`
+	}
+	USDValues := []*tokenUSDValue{}
+	if err := meddler.QueryAll(l2db.dbRead, &USDValues, query, args...); err != nil {
+		return tracerr.Wrap(err)
+	}
+
+	// Calculate average fee per transaction
+	var avgFeeUSD float64 = 0
```
Mikelle, line 159: Float64 by default is zero, so you can use just `var avgFeeUSD float64`

db/l2db/apiqueries.go
```diff
@@ -104,11 +105,123 @@ func (l2db *L2DB) AddTxAPI(tx *PoolL2TxWrite) error {
...
+	for _, USDValue := range USDValues {
+		avgFeeUSD += feeMap[USDValue.TokenID] * USDValue.USDPerToken
+	}
+	avgFeeUSD = avgFeeUSD / float64(len(txs))
```
Mikelle, line 163: I think it's better to use the decimal library (https://github.com/shopspring/decimal) for those types of calculations. Float64 is not suitable.

Because float64 (or any binary floating point type, actually) can't represent numbers such as 0.1 exactly.

Consider this code: http://play.golang.org/p/TQBd4yJe6B You might expect that it prints out 10, but it actually prints 9.999999999999831. Over time, these small errors can really add up!

arnaubennassar: Thanks for your suggestion, you're totally right. However I think this is a bit out of the scope of this PR as it would require a relatively big refactor:
- `func (f FeeSelector) Percentage() float64` should return `decimal.Decimal`
- The different types of txs (`Tx`, `PoolL2Tx`, `L2Tx` and `L1Tx`) have a field  `AmountFloat float64` which is used in some places to simplify the code (easier to use than `big` or `decimal` packages). This is also propagated to the DB.

In general we could say that there is no precision safeness across the repo, except when it's truly needed. Therefore I'd recommend to open an issue and tackle the problem from the root. Another related improvement that would emerge is getting rid of the `amount_f NUMERIC` columns in the database. This were used to do numeric calculations directly in SQL when we stored `big.Int` as bytes. However now they're stored as `DECIMAL(78,0)` so there's no need for that anymore.

If you agree, please mark the conversation as resolved and I'll create a detailed issue to do this refactor.

arnaubennassar: A part from that, I think that precision here is not that relevant:
- time between token value updates will be more relevant than precision issues
- This is used only to check if the fee falls between the threshold defined by users using the config file, it's kind of a sanity check. At the end the `txselector` is responsible for making decisions in terms of profitability

Mikelle (approved):

api/txspool.go
```diff
@@ -38,6 +40,79 @@ func (a *API) postPoolTx(c *gin.Context) {
...
+		retSQLErr(err, c)
+		return
+	}
+	// Return IDs of the added txs in the pool
+	c.JSON(http.StatusOK, txIDStrings)
+}
+
+// isAtomicGroup returns true if all the txs are needed to be forged
+// (all txs will be forged in the same batch or non of them will be forged)
+func isAtomicGroup(txs []l2db.PoolL2TxWrite) bool {
```
Pantani, line 85: I agree to use the graph package, but we can think about in use this nutshell implementation in the future:

https://github.com/hermeznetwork/hermez-node/blob/8c47d4f3a7b4dc3e46d23d16b0e8eb010a86144e/txselector/txbatch.go#L228-L293

Pantani: Can you create tests for this method?

api/txspool.go
```diff
@@ -38,6 +40,79 @@ func (a *API) postPoolTx(c *gin.Context) {
...
+	// Create a graph from the given txs to represent requests between transactions
+	g := graph.New(len(txs))
+	idToPos := make(map[common.TxID]int, len(txs))
+	// Map tx ID to integers that will represent the nodes of the graph
+	for i, tx := range txs {
+		idToPos[tx.TxID] = i
+	}
+	// Create vertices that connect nodes of the graph (txs) using RqTxID
+	for i, tx := range txs {
+		if tx.RqTxID == nil {
```
Pantani, line 95: whats happen if only one tx doesn't have a `RqTxID` parameter

api/txspool.go
```diff
@@ -38,6 +40,79 @@ func (a *API) postPoolTx(c *gin.Context) {
...
+	// Insert to DB
+	if err := a.l2.AddAtomicTxsAPI(writeTxs); err != nil {
+		retSQLErr(err, c)
+		return
+	}
+	// Return IDs of the added txs in the pool
+	c.JSON(http.StatusOK, txIDStrings)
+}
+
+// isAtomicGroup returns true if all the txs are needed to be forged
```
Pantani, line 83: Can you create tests for this method?

api/txspool.go
arnaubennassar: The idea of this endpoint is that all submitted txs are tomic within the same group (all the txs sended in the request have to be forged together). Therefore if only one tx doesn't request any other, this tx could be forged alone, making the group not 100% atomic.

Of course this is an arbitrary criteria that I made up, and we can change it. Let me share my reasoning:

- Someone could sneak txs pretending to be part of the atomic group using other people's fees to subvention their own txs (of course we could add logic to cover this case, but this way we avoid this extra logic)
- Prevent users from doing not intended things: atomic txs are not that easy to reason about, if we've an endpoint dedicated to atomic txs my guess is that people will send txs to this endpoint with the idea of this txs will all get forged or non of them will get forged. And this is exactly what I've implemented

Pantani (approved):

PR #825, opened 2021-05-20
Set a gasprice limit after receive it from external source

ARR552:
### What does this PR does?
This PR sets the max gasprice to 2000.

## Checklist

review them:

- [ ] Include tests

### DIFF
--- a/coordinator/txmanager.go
+++ b/coordinator/txmanager.go
@@ -161,6 +161,11 @@ func (t *TxManager) NewAuth(ctx context.Context, batchInfo *BatchInfo) (*bind.Tr
 			return nil, tracerr.Wrap(er)
 		}
 	}
+	//If gas price is higher than 2000, probably we are going to get the gasLimit exceed error
+	const masGasPrice = 2000
+	if gasPrice.Cmp(big.NewInt(masGasPrice)) == 1 {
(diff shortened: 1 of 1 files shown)


coordinator/txmanager.go
```diff
@@ -161,6 +161,11 @@ func (t *TxManager) NewAuth(ctx context.Context, batchInfo *BatchInfo) (*bind.Tr
 			return nil, tracerr.Wrap(er)
 		}
 	}
+	//If gas price is higher than 2000, probably we are going to get the gasLimit exceed error
+	const masGasPrice = 2000
```
jeffprestes, line 165: I guess there is a typo, the const name would be maxGasPrice. Is that correct?

jeffprestes: Also, as we already know that maxGasPrice constant will be compared to the gasPrice ( a big.Int value), what if we already define the constant as big.Int? This way it will save several cast later

arnaubennassar, line 165:
```suggestion
	const maxGasPrice = 2000
```

arnaubennassar (changes requested):

arnaubennassar (approved):

jeffprestes (approved):

PR #834, opened 2021-05-26
Accounts endpoint checks query params

### LINKED ISSUE #830: Invalid return at accounts' endpoint when an invalid parameter is sent
## Summary of Bug

When you send a request to the /accounts endpoint with an invalid parameter instead of return 400 as it is defined in REST protocol it is returning all accounts.

### Expected Behavior

If the parameter is invalid the server must return 400 and explain witch parameter is wrong or missing. This document gives some good example how to handle it https://developer.apple.com/documentation/appstoreconnectapi/interpreting_and_handling_errors/parsing_the_error_response_code

### Steps to Reproduce

https://api.internaltestnet.hermez.io/v1/accounts?hez:hez:25641

ARR552:
Closes #830

### What does this PR does?

This PR adds a validation in endpoint "v1/accounts/" that check if the query params used are correct. If not, the response code is 400 (Bad Request)

### How to test?

Use an incorrect query param in the endpoint "v1/accounts"

## Checklist

review them:

- [ ] Include tests

### DIFF
--- a/api/account.go
+++ b/api/account.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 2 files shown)


Pantani:
maybe we should use the gin binding and validation: https://github.com/gin-gonic/gin#model-binding-and-validation

jeffprestes (dismissed):

arnaubennassar (changes requested):
Please add test for this scenarios

arnaubennassar (approved):

Pantani (approved):

PR #839, opened 2021-05-27
set eth_tx_hash by default to 0x0

ARR552:
### What does this PR does?

This PR modifies the migration 3 and set the default value of the eth_tx_hash column to 0x0.

### How to test?

Run the migration and check the default value ofeth_tx_hash with the next query:
```
SELECT column_name, column_default
FROM information_schema.columns
WHERE (table_schema, table_name) = ('public', 'batch')
ORDER BY ordinal_position;
```

## Checklist

review them:

- [X] Include tests


arnaubennassar (approved):
sugeestion: instead of adding the `COALESCE` in the queries, add an `UPDATE` in the migration file to set `000...00` to columns that are `NULL`. It will take longer to run the migration (only for existing nodes, and only once) instead of affecting the performance of all the queries that we'll do from now on

jeffprestes:
It was my suggestion to add COALESCE in the query, this way we can avoid NULL error in the case something wrong happens in the migrations. It's a conservative and sounds as redundant check but after the issue we faced on testnet I think it's better we have a tiny performance lost in the query instead of an error. But it can be removed in the future code refactor.

Pantani (approved):

PR #841, opened 2021-05-28
Feature/reafactor pooll2tx types

arnaubennassar:
Partialy implements #833

### What does this PR does?

- With this refactor, sending pool txs from an external Go project is as easy as marshaling `common.PoolL2Tx` into a JSON 
- Merge structs Merge `api.receivedPoolTx`, `api.testPoolTxSend` and `l2db.PoolL2TxWrite` into `common.PoolL2Tx`
- Use the same method to insert txs into the pool for testing and API (production)
- Prevent method `func (tx *PoolL2Tx) RqTxCompressedDataV2() (*big.Int, error)` from modifying `tx.RqAmount` value 

Note that `common.formats.go` and `common.formats.go` (which represents about 70% of the added code) are basically copied code from `api/apitypes` and `api` packages. In future PRs I will finish the refactoring by deleting the equivalent code from `./api/**`. I didn't make it in this one since the amount of altered code would have made the PR very hard to reason about.

### How to test?

Since this is a refactor, existing tests should be valid

## Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Include tests
- [x] Respect code style and lint
- [ ] Update swagger file (if needed)
- [ ] Update documentation (*.md) (if needed)

PR #842, opened 2021-05-29
Feature/marshal json refactor

### LINKED ISSUE #801: MarshalJSON functions must use defined struct instead of interface{}
## Rationale

Today as you can see at https://github.com/hermeznetwork/hermez-node/blob/51d4dd5390df5f99eb746e315c751194fd700d37/db/historydb/views.go#L64 the Marshal function create a generic struct in runtime. This does not allow third-party software, such as a SDK or any other software, to reuse hermez-node code and assure the struct it is being used is the same used on hermez-node and keep the compatibility with confidence. 

Also, is not idiomatic to use interface{} in Go. Always is preferable to define a struct and in this case seems to do not have a reason to not use a defined struct.

This change must be applied in all MarshallJSON in historydb package

## Implementation

Create a struct for the data format is going to be returned by API instead of interface{}

mfcastellani:
Closes #801

## Rationale

Today as you can see at https://github.com/hermeznetwork/hermez-node/blob/51d4dd5390df5f99eb746e315c751194fd700d37/db/historydb/views.go#L64 the Marshal function create a generic struct in runtime. This does not allow third-party software, such as a SDK or any other software, to reuse hermez-node code and assure the struct it is being used is the same used on hermez-node and keep the compatibility with confidence. 

Also, is not idiomatic to use interface{} in Go. Always is preferable to define a struct and in this case seems to do not have a reason to not use a defined struct.

This change create structs and remove the interface use in the historydb package. 

### How to test?

`make test`

## Checklist

review them:

- [X] Include tests - since PR is just a refactor tests are already available
- [X] Update documentation (*.md) (if needed)

### DIFF
--- a/db/historydb/views.go
+++ b/db/historydb/views.go
@@ -64,8 +64,8 @@ type txAPIJSON struct {
 	BatchNum    *common.BatchNum     `json:"batchNum"`
 	HistoricUSD *float64             `json:"historicUSD"`
 	Timestamp   time.Time            `json:"timestamp"`
-	L1Info      l1infoJSON           `json:"L1Info"`
-	L2Info      l2infoJSON           `json:"L2Info"`
+	L1Info      *l1infoJSON          `json:"L1Info"`
+	L2Info      *l2infoJSON          `json:"L2Info"`
 	TokenJSON   tokenJSON            `json:"token"`
 	L1orL2      string               `json:"L1orL2"`
 }
@@ -143,10 +143,15 @@ func (tx TxAPI) MarshalJSON() ([]byte, error) {
 		FromIdx:     tx.FromIdx,
 		FromEthAddr: tx.FromEthAddr,
 		FromBJJ:     tx.FromBJJ,
+		ToIdx:       tx.ToIdx,
+		ToEthAddr:   tx.ToEthAddr,
+		ToBJJ:       tx.ToBJJ,
 		Amount:      tx.Amount,
 		BatchNum:    tx.BatchNum,
 		HistoricUSD: tx.HistoricUSD,
 		Timestamp:   tx.Timestamp,
+		L1Info:      nil,
+		L2Info:      nil,
 		TokenJSON: tokenJSON{
 			TokenID:          tx.TokenID,
 			TokenItemID:      tx.TokenItemID,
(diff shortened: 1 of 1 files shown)


db/historydb/views.go
```diff
@@ -12,6 +12,80 @@ import (
...
+	ToForgeL1TxsNum          *int64              `json:"toForgeL1TransactionsNum"`
+	UserOrigin               *bool               `json:"userOrigin"`
+	DepositAmount            *apitypes.BigIntStr `json:"depositAmount"`
+	AmountSuccess            bool                `json:"amountSuccess"`
+	DepositAmountSuccess     bool                `json:"depositAmountSuccess"`
+	HistoricDepositAmountUSD *float64            `json:"historicDepositAmountUSD"`
+	EthereumBlockNum         int64               `json:"ethereumBlockNum"`
+}
+
+type l2infoJSON struct {
```
jeffprestes, line 83: @mfcastellani , could you make these new structs public? This way we can also reuse them in SDK or external libraries.

jeffprestes (changes requested):
As asked in the issue, could you please make these new structs public? All other changes looks good to me.

arnaubennassar (dismissed):
I've been working in a similar refactor recently, and I wanted to share a few reflections and ideas on this topic (just to be clear, I'm not asking for adding this on this PR):

- We should add Unmarshalers and make structs public, to make it easy for other projects to integrate
- Once we've public structs that are able to marshal/unmarshal in the api format, they shouldn't use the `apitypes` types. Example: `EthAddr          apitypes.HezEthAddr` should be `EthAddr          ethCommon.Address`. Reasoning: until now we only cared to fulfill the API format, as it was the main duty, but this is very bad for reusing code on the client side, people don't want tor receive a weird string  but an actual Ethereum address (in this example, but idx, bjj, .... also aplies). To do so I suggest hiding this intermediary structs inside the marhsal/unmarshal functions [as I did here](https://github.com/hermeznetwork/hermez-node/pull/841/files#diff-6bb7ed027df678ce6822860d0344e836821a8c7e6ab979291dad26de7e48051aR444)
- We should try to use the `common` package as much as possible to avoid having similar structs that represent the same but require extra parsing and similar boring code. Although this is impossible in many cases due to the difference in content from the API objects and the common structs.
- I'd like to reconsider the overall architecture: to me it's a bit ugly to have things that are very specific to the api package in the db packages

tclemos:
> * I'd like to reconsider the overall architecture: to me it's a bit ugly to have things that are very specific to the api package in the db packages

I've seen problems like this before and having a project architecture well defined can help. 

The Hexagonal Architecture is an Architecture that aims to have the core domain isolated from the external dependencies like http, db, infrastructure in general, etc.

Please, take a look in this article and see if it helps: https://medium.com/@matiasvarela/hexagonal-architecture-in-go-cfd4e436faa3

jeffprestes (approved):

PR #857, opened 2021-06-04
Fix txselector L1CoordTxs

### LINKED ISSUE #866: Revert zki tests to match compatibility canary
## Rationale

- We should remain compatible with compatibility canary
- Once we're we will need to fix existing unit tests

arnaubennassar:
Close #866 

### What does this PR does?

- Revert the asserts of `test/zkproof/flows_test.go` so they match the ones in `compatibility-canary`
- When a coordinator creates accounts on behalf of the user, this transactions where grouped under `l1UserTxs` instead of `l1CoordTxs`, fixed it
- Added some `// TODO:` coments in part of the code that we should undertand better. They don't necessarily represent an actual request to change code, but just a pointer on stuff that we've to look deeper into it 

### How to test?

Unit tests now match `compatibility-canary` again

## Checklist

review them:

- [ ] Include tests

### DIFF
--- a/txselector/txselector.go
+++ b/txselector/txselector.go
@@ -122,6 +122,7 @@ func (s *TxSelector) getL1L2TxSelection(selectionConfig txprocessor.Config, l1Us
 	for _, tx := range l1UserTxs {
 		log.Debugw("TxSelector: processing L1 user tx", "TxID", tx.TxID.String())
 		// assumption: l1usertx are sorted by L1Tx.Position
+		// TODO: why is the output of the function ignored?
 		_, _, _, _, err := processor.ProcessL1Tx(nil, &tx) //nolint:gosec
 		if err != nil {
 			return nil, nil, nil, nil, nil, nil, tracerr.Wrap(err)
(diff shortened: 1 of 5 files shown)


txselector/txselector.go
```diff
@@ -122,6 +122,7 @@ func (s *TxSelector) getL1L2TxSelection(selectionConfig txprocessor.Config, l1Us
 	for _, tx := range l1UserTxs {
 		log.Debugw("TxSelector: processing L1 user tx", "TxID", tx.TxID.String())
 		// assumption: l1usertx are sorted by L1Tx.Position
+		// TODO: why is the output of the function ignored?
```
mfcastellani, line 125: From what I understand from the code these cases where the method output is ignored is when an error occurs, and then all parameters are returned as nil. My understanding is that this part is correct and can be kept that way, but I still lack a lot of business concept because I'm not even a month into the project, so my conclusion could be totally wrong.

arnaubennassar: After looking a bit into it, it seems that this method is called in other parts of the code where the output and input are not ignored. I believe this function is placed here in order to check if there is an error on the L1 txs and maybe it does some changes on the state. In any case I think that this is hard to understand, therefore I'd suggest to:

1. Understand exactly why is this method being called
2. Refactor a bit by
    a) Adding a description on what this does and why (go deeper than `// Process L1UserTxs`)
    b) Split the logic of the function `ProcessL1Tx` and give meaningful names

jeffprestes (approved):

mfcastellani (approved):

PR #870, opened 2021-06-14
Feature/merge develop with atomic tx branch

mfcastellani:
This is a utility PR. I already fixed the conflicts in the code and just need @arnaubennassar approvation.

One important point, we have 2 migrations called 0004.sql, so since the 0004.sql from develop was already in develop I moved the 0004.sql previously in atomic-tx branch to 0005.sql. This maybe can be a issue to update my node, but is the correct path to follow.


jeffprestes (approved):

db/migrations/0005_test.go
```diff
@@ -0,0 +1,195 @@
...
+	assert.NoError(t, row.Scan(&result))
+	assert.Equal(t, 1, result)
+	// check that rq_tx_id colum doesn't exist anymore
+	const queryCheckItemID = `SELECT COUNT(*) FROM tx_pool WHERE rq_tx_id IS NULL;`
+	row = db.QueryRow(queryCheckItemID)
+	assert.Equal(t, `pq: column "rq_tx_id" does not exist`, row.Scan(&result).Error())
+}
+
+func TestMigration0005(t *testing.T) {
+	runMigrationTest(t, 4, migrationTest0005{})
```
arnaubennassar, line 194:
```suggestion
	runMigrationTest(t, 5, migrationTest0005{})
```

arnaubennassar (changes requested):

arnaubennassar (approved):

PR #883, opened 2021-06-18
Feature/check balance

mfcastellani:
### What does this PR does?

This PR add logic to validate balance inside splitL2ForgableAndNonForgable

We will need a simulator to check if all transactions are possible, currently if I have this scenario:

`UserA has 1000 in balance and try to send 1000 to UserB and UserC`

Code will fail in avoid this. We need a simulator (as I talked with @jeffprestes today).

## Checklist

review them:

- [ ] Include tests

### DIFF
--- a/txselector/txselector.go
+++ b/txselector/txselector.go
@@ -130,7 +130,7 @@ func (txsel *TxSelector) GetL2TxSelection(selectionConfig txprocessor.Config, l1
 	[][]byte, []common.L1Tx, []common.PoolL2Tx, []common.PoolL2Tx, error) {
 	metric.GetL2TxSelection.Inc()
 	coordIdxs, accCreationAuths, _, l1CoordinatorTxs, l2Txs,
-	discardedL2Txs, err := txsel.getL1L2TxSelection(selectionConfig,
+		discardedL2Txs, err := txsel.getL1L2TxSelection(selectionConfig,
 		[]common.L1Tx{}, l1UserFutureTxs)
 	return coordIdxs, accCreationAuths, l1CoordinatorTxs, l2Txs,
 		discardedL2Txs, tracerr.Wrap(err)
@@ -149,7 +149,7 @@ func (txsel *TxSelector) GetL1L2TxSelection(selectionConfig txprocessor.Config,
 	[]common.L1Tx, []common.PoolL2Tx, []common.PoolL2Tx, error) {
 	metric.GetL1L2TxSelection.Inc()
 	coordIdxs, accCreationAuths, l1UserTxs, l1CoordinatorTxs, l2Txs,
-	discardedL2Txs, err := txsel.getL1L2TxSelection(selectionConfig, l1UserTxs, l1UserFutureTxs)
+		discardedL2Txs, err := txsel.getL1L2TxSelection(selectionConfig, l1UserTxs, l1UserFutureTxs)
 	return coordIdxs, accCreationAuths, l1UserTxs, l1CoordinatorTxs, l2Txs,
 		discardedL2Txs, tracerr.Wrap(err)
 }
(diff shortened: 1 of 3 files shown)


txselector/txselector.go
```diff
@@ -782,6 +782,18 @@ func setAtomicMetadata(tp *txprocessor.TxProcessor, l2Txs []common.PoolL2Tx) (
...
+		enoughBalance, balance, feeAndAmount := tp.CheckEnoughBalance(l2Txs[i])
+		if !enoughBalance {
+			// not valid Amount with current Balance. Discard L2Tx,
+			// and update Info parameter of the tx, and add it to
+			// the discardedTxs array
+			l2Txs[i].Info = fmt.Sprintf("Tx not selected due to not enough Balance at the sender. "+
+				"Current sender account Balance: %s, Amount+Fee: %s",
+				balance.String(), feeAndAmount.String())
+			continue
+		}
```
arnaubennassar, line 795: The balance check is performed later on, I think we shouldn't do this here. Apart from that, the `Info` message doesn't have any effect since it's not appended to any of the returning slices. The idea is that this function splits txs in two groups:

- l2TxsForgable: txs that can be forged immediately or in future iterations
- invalidL2Txs: txs that can NOT be forged in any iteration. This could be due to multiple reasons such as:  nonce is smaller than the current nonce of the sender account, transfer to idx and destinatary doesn't exists malformed atomic group.

For the case of balance, is not enough to assume that the tx is not valid as other txs could send funds to the account, making this tx forgable in the same batch. As a mater of fact, the test that you've edited needs more batches to forge the same amount of txs

arnaubennassar (changes requested):

txselector/txselector.go
jeffprestes: ```the Info message doesn't have any effect since it's not appended to any of the returning slices.``` this is true.
```For the case of balance, is not enough to assume that the tx is not valid as other txs could send funds to the account, making this tx forgable in the same batch.``` sorry this sentence doesn't seems be talking about balance. Could you be more specific @arnaubennassar ? 
So far, I agree with Marcelo, the first balance check can be applied here as any other verification. I suggest you @mfcastellani to keep it and test.

jeffprestes (commented):
Small changes needed to be implemented

txselector/txselector.go
arnaubennassar: This function is only run once (before going to the actual selection process), and it's purpose is to set the atomic related metadata for the selection process to be aware of which txs are atomic, and to know which atomic txs are in the same groups. 

Since there can be malformed atomic groups it's convenient to allow the function to discard txs that are part of malformed atomic groups. Additionally, the function could also filter txs that won't be forgable in the batch, the reasons for discarding txs could be:

- Nonce is lower than the current account nonce (it's impossible for the nonce to go backwards compared to the starting point)
- Transfer to idx where the idx doesn't exist, or the TokenID doesn't match (although idx can be created after this point, it's impossible for users to know to whom those idxs will be assigned, so it's kind of user safety feature to discard txs in this situation)

Any other reason could became valid due to other txs altering the state. Since the txs discarded by this function will be ignored until next batch processing, discarding because of balance at this step causes a sub optimal selection (this can be clearly seen in the test that is edited in this PR). Furthermore, balance MUST be checked later on, so it's safe to not do this check here

arnaubennassar: Simple explanation: this functions should only discard txs that will not be forgable in the given batch, without taking into consideration other txs

mfcastellani: Checking the flow, the points that @arnaubennassar mentioned here make sense. The balance filtering that occurs later is implemented today in a generic way (it occurs for atomic transactions and for non-atomic transactions), if it is added here we will have a duplication of the same logic, and in this case it is inefficient for an atomic block.

The point is: this validation that takes place for individual transactions does not work for the concept of atomic transactions. Imagine the hypothetical situation where user A has 1000 on balance and wants to transfer 1000 to user B and 1000 to user C. This group of transactions will fail at the time of completion.

To solve this I believe that we should have a kind of simulator at this point in the code validating not the transactions, but the transaction groups in order to predict if it is possible to perform the entire group of transactions before sending them forward.

arnaubennassar:
Please close this PR as it's no longer relevant

PR #889, opened 2021-06-22
Feature/atomic tx api

### LINKED ISSUE #886: Simplify `POST /atomic-pool`, and add `RqOffset`
## Rationale

This feature request considerates to items:

- [x] There is no need to include the `request*` fields on the JSON request of the endpoint since this information is already as part of the transaction that is being requested. So we should remove all of them, except the `requestId`, which is the one that will help us find the tx that we need to set the actual `Rq` fields of the Go struct.
- [x] Add the `requestOffset` as part of the JSON request of the endpoint. This will make the user able of choosing the actual order in which transactions will be processed. (Note that this cannot be enforced in terms of signature, an therefore a malitious coordinator could alter the order of the transactions)

**EDIT:** for safety reasons, also:

- [ ] Reject any transaction in `POST /transactions-pool` that include `requestOffset` or `requestId` fields in the JSON request.

## Implementation

- Edit the JSON marshal/umnarshal interface of `common.PoolL2Tx`
- Edit the swagger documentation
- Add `rq_offset` column in `tx_pool` table
- Modify the function that inserts txs into the table so it also inserts the `RqOffset`
- Once the JSON request has been received an unmarshalled (`api/txspool.go:postAtomicPool`), iterate over the array, and set the `Rq*` fields (except for `RqTxID` and `RqOffset`, which are the ones that should have been set coorectly by the user). Tip: the RqOffset could be used to find the requested item without needing to iterate over the entire array. If the `RqOffset` hasn't been set properly by the user the endpoint should return an error anyway. **This has to be done before `verifyPoolL2Tx`, since this function expects the fields to be already set in order to verify the signature.**

```go
// Consider this pseudo code, I don't even know if this compiles

for i := 1; i < len(txs); i++ {
	if txs[i].RqTxID == common.EmptyTxID // ... return error (only atomic txs allowed)
        // validate that txs[i].RqOffset is valid in terms of allowed values as per protocol
        if txs[i + txs[i].RqOffset] != txs[i].RqTxID // ...return error: malformed atomic group (it could be bad order or missing tx)
        requestedTx := txs[i + txs[i].RqOffset]
        txs[i].RqFromIdx = requestedTx.FromIdx
        // Do the same with remaining Rq* fields
}
```

**Considerations about JSON format:** RqOffset should follow the protocol format, as this is not very complicated, and could be a source of problems

**EDIT II:** to help the `txselector`, a new field should be added to `common.PoolL2Tx`: `AtomicGroupID`. This field should have a unique value per group and should be added in the DB as well as the `rq_offset`. It can be implemented in many ways, for instance using `uuid` or by hashing the concatenation of the different `RqTxID` of the transactions that conform the atomic group

mfcastellani:
Related to #886

### What does this PR does?

In order to simplify POST with atomic txs this PR remove some fields from POST /transactions-pool JSON

### How to test?

make test

## Checklist

review them:

- [X] Include tests


api/txspool_test.go
```diff
@@ -127,29 +119,6 @@ func genTestPoolTxs(
 			bjj := bjjToString(acc.BJJ)
 			genReceiveTx.ToBJJ = &bjj
 		}
-		if poolTxs[i].RqFromIdx != 0 {
-			rqFromIdx := idxToHez(poolTxs[i].RqFromIdx, rqToken.Symbol)
-			genReceiveTx.RqFromIdx = &rqFromIdx
-			genReceiveTx.RqTxID = &poolTxs[i].RqTxID
```
arnaubennassar, line 133: RqTxID should be kept

common/pooll2tx.go
```diff
@@ -523,32 +513,6 @@ func (tx PoolL2Tx) MarshalJSON() ([]byte, error) {
 		toBJJ := bjjToString(tx.ToBJJ)
 		toMarshal.ToBJJ = &toBJJ
 	}
-	// Set Rq fields
-	if tx.RqFromIdx != 0 {
-		if tx.RqTokenSymbol == "" {
-			return nil, errors.New("Invalid tx.RqTokenSymbol")
-		}
-		toMarshal.RqTxID = &tx.RqTxID
```
arnaubennassar, line 531: RqTxID should be kept

arnaubennassar (changes requested):
Please also add `AtomicGroupID` into `common.PoolLTx`. You don't need to make anything with it yet, but it will be handy for me to have this ASAP

arnaubennassar:
Like the approach of this PR in terms of small scope. I believe you're just missing two things to pass the test:

- Include the RqTxID in the JSON response
- Set the `Rq` fields in `POST /atomic-pool`, you'll need to iterate over the array to find the linked txs to set this Rq fields. Note that this has to be done in order to verify the signature

api/txspool.go
```diff
@@ -47,6 +47,27 @@ func (a *API) postAtomicPool(c *gin.Context) {
...
+				tx1.RqToEthAddr = tx2.ToEthAddr
+				tx1.RqToBJJ = tx2.ToBJJ
+				tx1.RqTokenID = tx2.TokenID
+				tx1.RqAmount = tx2.Amount
+				tx1.RqFee = tx2.Fee
+				tx1.RqNonce = tx2.Nonce
+				break
+			}
+			// check if was last and not set
+			if i == (len(receivedTxs) + 1) {
```
arnaubennassar, line 65:
```suggestion
			if i == (len(receivedTxs) - 1) {
```

api/txspool.go
```diff
@@ -47,6 +47,27 @@ func (a *API) postAtomicPool(c *gin.Context) {
 		retBadReq(errors.New(ErrSingleTxInAtomicEndpoint), c)
 		return
 	}
+	// set the Rq fields
+	for _, tx1 := range receivedTxs {
+		for i, tx2 := range receivedTxs {
+			if tx1.RqTxID == tx2.TxID {
+				tx1.RqFromIdx = tx2.FromIdx
```
arnaubennassar, line 54: I'd suggest to not use this pattern when writing to the items you're iterating over, since some times there is unexpected pointer related behavior.

Instead I'd do something like:

`receivedTxs[i].RqFromIdx = receivedTxs[j].FromIdx`

common/pooll2tx.go
```diff
@@ -557,27 +554,19 @@ func (tx PoolL2Tx) MarshalJSON() ([]byte, error) {
...
+		FromIdx   StrHezIdx             `json:"fromAccountIndex" binding:"required"`
+		ToIdx     StrHezIdx             `json:"toAccountIndex"`
+		ToEthAddr StrHezEthAddr         `json:"toHezEthereumAddress"`
+		ToBJJ     StrHezBJJ             `json:"toBjj"`
+		Amount    *StrBigInt            `json:"amount" binding:"required"`
+		Fee       FeeSelector           `json:"fee"`
+		Nonce     Nonce                 `json:"nonce"`
+		Signature babyjub.SignatureComp `json:"signature" binding:"required"`
+		RqTxID    TxID                  `json:"requestId"`
+		State     string                `json:"state"`
```
arnaubennassar, line 569: Missing `RqOffset`

common/pooll2tx.go
```diff
@@ -592,29 +581,20 @@ func (tx *PoolL2Tx) UnmarshalJSON(data []byte) error {
...
+		ToBJJ:       babyjub.PublicKeyComp(receivedJSON.ToBJJ),
+		TokenID:     receivedJSON.TokenID,
+		Amount:      (*big.Int)(receivedJSON.Amount),
+		Fee:         receivedJSON.Fee,
+		Nonce:       receivedJSON.Nonce,
+		Signature:   receivedJSON.Signature,
+		RqTxID:      receivedJSON.RqTxID,
+		Type:        receivedJSON.Type,
+		State:       state,
+		TokenSymbol: receivedJSON.FromIdx.TokenSymbol,
```
arnaubennassar, line 597: Missing `RqOffset`

common/pooll2tx.go
```diff
@@ -523,7 +521,6 @@ func (tx PoolL2Tx) MarshalJSON() ([]byte, error) {
 		toBJJ := bjjToString(tx.ToBJJ)
 		toMarshal.ToBJJ = &toBJJ
 	}
-	// Set Rq fields
 	if tx.RqFromIdx != 0 {
 		if tx.RqTokenSymbol == "" {
```
arnaubennassar, line 525: Should remove all the fields except `RqTxID`, and add `RqOffset`

arnaubennassar (changes requested):

arnaubennassar (approved):

Issue #859, opened 2021-06-07
Monitor balance of the coordinator


arnaubennassar:
## Rationale

Include the coordinator Ether balance as part of the `GET /health` response. This will help potential monitoring systems to prevent forging fails due to insufficient balance.

PR #898, opened 2021-06-24
added coordinator balance to health endpoint

Mikelle:
Closes https://github.com/hermeznetwork/hermez-node/issues/859

### What does this PR does?

Added coordinator balance to health endpoint

### How to test?

Run request to /v1/health to a node in coord mode.

## Checklist

review them:

- [ ] Include tests

### DIFF
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -3457,7 +3457,7 @@ components:
           type: string
           description: Current version of the hermez-node
           example: v1.2.0
-        coordinatorBalance:
+        coordinatorForgerBalance:
           type: string
           description: Coordinator balance in ether. Returned with 18 decimals.
           example: 48901270356066543960
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -3457,6 +3457,10 @@ components:
           type: string
           description: Current version of the hermez-node
           example: v1.2.0
+        coordinatorBalance:
+          type: string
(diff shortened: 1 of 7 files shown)


api/swagger.yml
```diff
@@ -3457,6 +3457,10 @@ components:
           type: string
           description: Current version of the hermez-node
           example: v1.2.0
+        coordinatorBalance:
```
arnaubennassar, line 3460: Coordinators can use two addresses: one for forging and one for bidding. Most of the cases this address is the same, but not all the times. I'd suggest to:

a) Specify in the swagger doc that this address refers to the one used for forging (and probably change the name of `coordinatorBalance`) to help users understand
b) Add the bidder address as well

arnaubennassar (dismissed):
Approving, as it's good enough. But please, consider implementing one of the two suggestion I've made

arnaubennassar:
Also, could you add tests for the endpoint?

Mikelle:
> Also, could you add tests for the endpoint?

Unfortunately, to test this field, we have to run node in coord mode and specify coordinator forger address. 
Attaching screenshot from such node, to see, that coordinatorForgerAddress is returning.
Screen Shot 2021-06-25 at 1 48 10 PM

mfcastellani (approved):

arnaubennassar (approved):

PR #914, opened 2021-06-29
Inprove the format of POST /atomic-poolby adding deterministic Atomic…

### LINKED ISSUE #905: Calculate atomic group ID in a deterministic way

### LINKED ISSUE #913: Modify API format for atomic transactions
## Rationale

Improve the API format for `POST /atomic-pool`

## Implementation

Depends on #905 

Implement this format:
```json
{
     "atomicGroupId": "0x...", // final format will be decided after closing #905, but it's very likely to be a hash just as the normal id for txs
     "transactions": [
          {
              " id": "0x....",
               // other fields, same as regular transactions
               "requestOffset": 1
          },
          {
               "id": "0x....",
               // other fields, same as regular transactions
               "requestOffset": 7
          }
     ]
}
```

As a reminder we will be following [this spec](https://docs.hermez.io/#/developers/protocol/hermez-protocol/circuits/circuits?id=rq-tx-verifier) for the `requestOffset` values

arnaubennassar:
Closes #913 
Closes #905 

### What does this PR does?

- Calculate the atomic group ID in a deterministic way, this change forces to change the type from int to bytes, and because of that, many lines on the code has changed
- Change the API request body format for `POST /atomic-pool`
- Completely remove `RqTxID` from all the code since it's not safe to use it to reference other txs 

### How to test?

Included unit tests

## Checklist

review them:

- [x] Include tests
- [x] Update swagger file (if needed)

### DIFF
--- a/api/txspool.go
+++ b/api/txspool.go
@@ -41,8 +41,8 @@ func (a *API) postPoolTx(c *gin.Context) {
 
 // AtomicGroup represents a set of atomic transactions
 type AtomicGroup struct {
-	AtomicGroupID common.AtomicGroupID `json:"atomicGroupId"`
-	Txs           []common.PoolL2Tx    `json:"transactions"`
+	ID  common.AtomicGroupID `json:"atomicGroupId"`
+	Txs []common.PoolL2Tx    `json:"transactions"`
 }
 
 // SetAtomicGroupID set the atomic group ID for an atomic group that already has Txs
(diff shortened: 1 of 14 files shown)


api/txspool.go
```diff
@@ -40,61 +39,94 @@ func (a *API) postPoolTx(c *gin.Context) {
 	c.JSON(http.StatusOK, receivedTx.TxID.String())
 }
 
+// AtomicGroup represents a set of atomic transactions
+type AtomicGroup struct {
+	AtomicGroupID common.AtomicGroupID `json:"atomicGroupId"`
```
Mikelle, line 44: Isn’t it better to name `AtomicGroupID` as `ID`, bcs from the code it's understandable, that this ID belongs to AtomicGroup?

Mikelle (approved):

api/txspool.go
arnaubennassar: Applied the change request

mfcastellani (approved):

PR #924, opened 2021-07-02
allows don't specify symbols and addresses in config file

ARR552:
### What does this PR does?

This PR allows don't specify symbols and addresses in the config file. Now, you can even remove those properties if you are not gonna use them.

### How to test?

Sync the hermez node in mainnet and check using config file without symbols and addresses properties.

## Checklist

review them:

- [X] Include tests

### DIFF
--- a/priceupdater/priceupdater.go
+++ b/priceupdater/priceupdater.go
@@ -111,6 +111,8 @@ func (d *symbolsMap) strToMapSymbol(str string) error {
 		}
 		d.Symbols = mapping
 		log.Debug("Symbol mapping from config file: ", mapping)
+	} else {
+		d.Symbols = make(map[uint]string)
 	}
 	return lastErr
 }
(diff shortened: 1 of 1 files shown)


priceupdater/priceupdater.go
```diff
@@ -331,15 +335,18 @@ func (p *PriceUpdater) UpdateTokenList() error {
...
 					provider.SymbolsMap.Symbols[dbToken.TokenID] = dbToken.Symbol
 				}
-			}
-			if len(provider.AddressesMap.Addresses) != 0 {
+			case UpdateMethodTypeCoingeckoV3:
 				if _, ok := provider.AddressesMap.Addresses[dbToken.TokenID]; !ok {
 					provider.AddressesMap.Addresses[dbToken.TokenID] = dbToken.Addr
 				}
+			default:
+				log.Error("Unknown provider detected: ", provider.Provider)
```
jeffprestes, line 348: Here I would suggest to log an info instead of error. Why? Because maybe the hermez node administrator intentionally  try to change to another price service and would be good to us to inform her that we don't support it. 
```log.Warn("This price service provider is not supported: ", provider.Provider)```

Mikelle (approved):

arnaubennassar (approved):

PR #927, opened 2021-07-02
Fix/remove swagger fiat

ARR552:
### What does this PR does?

Remove the fiat endpoint from the swagger file

## Checklist

review them:

- [ ] Include tests


jeffprestes (dismissed):

arnaubennassar:
Could you add tests for this endpoint? I believe it's OK to ignore tests on the release, since we know they're passing and the code is not gonna change. However I think it's important we have them on `develop`

arnaubennassar:
Also, thanks for including my commits that I added on the release branch but not here, I forgot to do it.

arnaubennassar (approved):

PR #946, opened 2021-07-06
Set more detailed info to non-selected atomic txs

### LINKED ISSUE #903: Add better logs for failed atomic txs
## Rationale

Set good tx.Info for failing atomic txs

arnaubennassar:
Closes #903 

### What does this PR does?

- Add info explaining why an atomic tx is not selected. It differentiates the tx that made the entire group fail from the others.

## Checklist

review them:

- [x] Include tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,6 +30,7 @@ require (
 	github.com/stretchr/testify v1.7.0
 	github.com/tyler-smith/go-bip39 v1.1.0 // indirect
 	github.com/urfave/cli/v2 v2.3.0
+	github.com/yourbasic/graph v0.0.0-20210606180040-8ecfec1c2869
 	go.uber.org/zap v1.16.0
 	golang.org/x/crypto v0.0.0-20210616213533-5ff15b29337e
 	golang.org/x/net v0.0.0-20210226172049-e18ecbb05110
--- a/txselector/txselector.go
+++ b/txselector/txselector.go
@@ -144,6 +144,12 @@ func (txsel *TxSelector) GetL1L2TxSelection(selectionConfig txprocessor.Config,
 		discardedL2Txs, tracerr.Wrap(err)
 }
 
+type failedAtomicGroup struct {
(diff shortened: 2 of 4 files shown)


mfcastellani (approved):

txselector/txselector.go
```diff
@@ -144,6 +144,12 @@ func (txsel *TxSelector) GetL1L2TxSelection(selectionConfig txprocessor.Config,
 		discardedL2Txs, tracerr.Wrap(err)
 }
 
+type failedAtomicGroup struct {
```
tclemos, line 147: since this struct is meant to contain only one transaction from the atomic group, I suggest calling this as `failedAtomicTransaction`

go.mod
```diff
@@ -30,10 +30,11 @@ require (
 	github.com/stretchr/testify v1.7.0
 	github.com/tyler-smith/go-bip39 v1.1.0 // indirect
 	github.com/urfave/cli/v2 v2.3.0
+	github.com/yourbasic/graph v0.0.0-20210606180040-8ecfec1c2869
```
tclemos, line 33: this is an import used on `txspool.go`, but this PR is not affecting this specific file, so I'm not sure why this is being added at this moment, anyway, I just wanted to point this out because it is something weird to me.

tclemos (approved):
I'm approving it because I understand none of my comments will affect the functionality itself, just the way we make the code expressive and double-checking imports, also this is a merge to a WIP feature branch, saying that I understand we can handle these suggestions in the future.

txselector/txselector.go
arnaubennassar: The `txID` is used to indicate the tx that made the entire group of txs fail. But the names I choose are not self explainatory, so I'll update this

go.mod
arnaubennassar: Yup it's wierd xD
I believe that at some point I didn't propagated something well, or made a branch with uncommited changes or something like that

Issue #846, opened 2021-05-31
Min gas price value as parameter


jeffprestes:
## Rationale

Today the min gas price is a constant, 5 gwei. It would be good this value could be get from a parameter in cfg.builder.toml file

## Implementation
```toml
MinGasPrice = 6000000000
```

PR #949, opened 2021-07-06
added min gas price as a parameter in the config

Mikelle:
Closes https://github.com/hermeznetwork/hermez-node/issues/846

### What does this PR does?

Adding min gas price as a parameter in the config

### How to test?

Run node with new config parameter in coord mode and try to forge transactions

## Checklist

review them:

- [ ] Include tests

### DIFF
--- a/coordinator/txmanager.go
+++ b/coordinator/txmanager.go
@@ -133,14 +133,10 @@ func (t *TxManager) NewAuth(ctx context.Context, batchInfo *BatchInfo) (*bind.Tr
 	// First we try getting the gas price from etherscan. Later we get the gas price from the ethereum node.
 
 	var err error
-
 	// If gas price is higher than 2000, probably we are going to get the gasLimit exceed error
 	const maxGasPrice = 2000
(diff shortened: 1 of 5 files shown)


coordinator/txmanager.go
```diff
@@ -161,7 +157,7 @@ func (t *TxManager) NewAuth(ctx context.Context, batchInfo *BatchInfo) (*bind.Tr
 	}
 
 	maxGasPriceBig := big.NewInt(maxGasPrice)
```
mfcastellani, line 159: Not related to this PR, but I did not find where the maxGasPrice value from config file, also I checked with @tclemos and @jeffprestes and we agreed with this. The value for maxGasPrice at the config file is not used.

Is possible to add to this PR in this line this suggestion and test? Or do you believe is better to create another PR to fix this @Mikelle? 

```suggestion
	maxGasPriceBig := big.NewInt(t.cfg.MaxGasPrice)
```

arnaubennassar (approved):

mfcastellani (approved):

arnaubennassar (approved):

PR #961, opened 2021-07-12
Dont load migrations on init db package

arnaubennassar:
Closes #<issue number>. _in case of a bug fix, this should point to a bug or any other related issue(s)_

### What does this PR does?

Don't load the migration data for SQL on `init()` function, as there are some scenarios in which this package could be imported directly or indirectly without the need or desire of having the actual SQL DB in place

## Checklist

review them:

- [ ] Include tests

### DIFF
--- a/db/utils.go
+++ b/db/utils.go
@@ -37,23 +37,26 @@ const (
 	OrderDesc = "DESC"
 )
 
-var migrations *migrate.PackrMigrationSource
-
-func init() {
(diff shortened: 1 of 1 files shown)


jeffprestes:
@arnaubennassar without it the packr will not find migrations. 
What is needed is do not have public struct that will be used by other app in db package. 
I would recommend to close this issue.

PR #963, opened 2021-07-12
Feature/txselector maxbatchnum

arnaubennassar:
### What does this PR does?

- Filter transactions when `MaxNumBatch > current batch num` at `txselector`
- Add zki signal for MaxNumBatch at `txprocessor`
- Purge transactions that became invalid due to `MaxNumBatch` at `db/l2db` (aka transaction pool)
- Added myself as owner for `txselector`

## Checklist

review them:

- [x] Include tests

### DIFF
--- a/txselector/txselector.go
+++ b/txselector/txselector.go
@@ -441,6 +441,7 @@ func (txsel *TxSelector) processL2Txs(
 	// TODO: differentiate between nonSelectedL2Txs and unforjableL2Txs
 	// (right now all fall into nonSelectedL2Txs, which is safe but non optimal)
 	positionL1 := nAlreadyProcessedL1Txs
+	nextBatchNum := uint32(txsel.localAccountsDB.CurrentBatch()) + 1
 	// Iterate over l2Txs
 	// - check Nonces
 	// - check enough Balance for the Amount+Fee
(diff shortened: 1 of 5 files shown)


txselector/txselector.go
```diff
@@ -472,10 +472,31 @@ func (txsel *TxSelector) processL2Txs(
 			break
 		}
 
+		// Check if MaxNumBatch is exceeded
+		if l2Txs[i].MaxNumBatch > uint32(txsel.localAccountsDB.CurrentBatch())+1 {
```
jeffprestes, line 476: Maybe here it would be:
```if l2Txs[i].MaxNumBatch < uint32(txsel.localAccountsDB.CurrentBatch())+1 { ```
because if the maximum batch number user wanted is 100 and current batch is 101 so it is become invalid. User wants his transaction processed before the current batch so if MaxNumBatch is 100 and the `uint32(txsel.localAccountsDB.CurrentBatch())+1` is 101 so the transaction is invalid.

arnaubennassar: It may be a bit confusing because of the comment, but if the expression is evaluated to true the tx is rejected.
Also `txsel.localAccountsDB.CurrentBatch()` return the last processed batch, but here we're running the selection for the next batch, that's the reason for the `+1`

mfcastellani (approved):

tclemos (approved):

PR #1000, opened 2021-07-16
Fix scientific notation in state root

### LINKED ISSUE #993: State root should be represented as a string in GET /health
## Rationale

The State Root is represented as a `big.Int`, however when this is transformed into JSON, such a big number will have precission problems and be displayed as something like: `2.0723800499378798e+76`. Since this value is a hash, it's completley useless if it's not 100% accurate

## Implementation

Transform big.Int into string

ARR552:
Closes #993

### What does this PR does?

This PR represents the state root as a string in health endpoint.

### How to test?

Run the hermez node and call the healh endpoint.

## Checklist

review them:

- [ ] Include tests

### DIFF
--- a/health/checkers/statedb.go
+++ b/health/checkers/statedb.go
@@ -35,7 +35,7 @@ func (sdb StateDBChecker) Check() health.Health {
 
 	h.Up().
 		AddInfo("batchNum", batchNum).
-		AddInfo("root", root)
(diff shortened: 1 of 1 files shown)


Mikelle (approved):

marioidival (approved):
LGTM

PR #1015, opened 2021-07-20
Check atomic groups integrity

### LINKED ISSUE #909: Check atomic group integrity in txselector
## Rationale

Right now, the `txselector` asumes that the atomic groups received from the pool are correct in terms of `RqOffset` and relative position. However there are many things that could go wrong in the DB (`external_delete`, purger deleting only one of the txs, ...) and it would be nice to have a extra security step to ensure that all requested txs are on the correct place.

## Implementation

While selecting txs, make sure that the requested tx exists in the correct position before processing it.

arnaubennassar:
Closes #909 
## Checklist

review them:

- [ ] Include tests

### DIFF
--- a/txselector/errors.go
+++ b/txselector/errors.go
@@ -67,9 +67,9 @@ const (
 	ErrNoAvailableSlotsType string = "ErrNoAvailableSlots"
 
 	// ErrInvalidAtomicGroup error message returned if an atomic group is malformed
-	ErrInvalidAtomicGroup = "Tx not selected because it belongs to an atomic group with missing transactions or bad requested tranaction"
+	ErrInvalidAtomicGroup = "Tx not selected because it belongs to an atomic group with missing transactions or bad requested transaction"
 	// ErrInvalidAtomicGroupCode error code
-	ErrInvalidAtomicGroupCode int = 17
+	ErrInvalidAtomicGroupCode int = 18
 	// ErrInvalidAtomicGroupType error type
-	ErrInvalidAtomicGroupType string = "ErrNoAvailableSlots"
+	ErrInvalidAtomicGroupType string = "ErrInvalidAtomicGroup"
 )
--- a/txselector/errors.go
+++ b/txselector/errors.go
@@ -65,4 +65,11 @@ const (
 	ErrNoAvailableSlotsCode int = 17
 	// ErrNoAvailableSlotsType error type
 	ErrNoAvailableSlotsType string = "ErrNoAvailableSlots"
+
+	// ErrInvalidAtomicGroup error message returned if an atomic group is malformed
+	ErrInvalidAtomicGroup = "Tx not selected because it belongs to an atomic group with missing transactions or bad requested tranaction"
+	// ErrInvalidAtomicGroupCode error code
+	ErrInvalidAtomicGroupCode int = 17
+	// ErrInvalidAtomicGroupType error type
+	ErrInvalidAtomicGroupType string = "ErrNoAvailableSlots"
 )
--- a/txselector/txselector.go
+++ b/txselector/txselector.go
@@ -6,7 +6,7 @@ This constrains can be splitted in two categories:
 Batch constrains (this information is passed to the txselector as `selectionConfig txprocessor.Config`):
 - NLevels: limit the amount of accounts that can be created by NLevels^2 -1.
 Note that this constraint is not properly checked, if this situation is reached the entire selection will fail
-- MaxFeeTx: maximum amount of different coordinator accounts (idx) thah can be used to collect fees.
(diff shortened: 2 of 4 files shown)


txselector/errors.go
```diff
@@ -65,4 +65,11 @@ const (
 	ErrNoAvailableSlotsCode int = 17
 	// ErrNoAvailableSlotsType error type
 	ErrNoAvailableSlotsType string = "ErrNoAvailableSlots"
+
+	// ErrInvalidAtomicGroup error message returned if an atomic group is malformed
+	ErrInvalidAtomicGroup = "Tx not selected because it belongs to an atomic group with missing transactions or bad requested tranaction"
```
mfcastellani, line 70:
```suggestion
	ErrInvalidAtomicGroup = "Tx not selected because it belongs to an atomic group with missing transactions or bad requested transaction"
```

txselector/txselector.go
```diff
@@ -1121,6 +1125,94 @@ func filterFailedAtomicGroups(
 	return txsToProcess, filteredTxs
 }
 
+// filterInvalidAtomicGroups split the txs into the ones that can be porcessed
```
mfcastellani, line 1128:
```suggestion
// filterInvalidAtomicGroups split the txs into the ones that can be processed
```

mfcastellani (changes requested):

mfcastellani:
LGTM, just two typos...

txselector/txselector.go
```diff
@@ -1121,6 +1125,94 @@ func filterFailedAtomicGroups(
...
+		atomicGroupID := txs[i].AtomicGroupID
+		if atomicGroupID == common.EmptyAtomicGroupID {
+			// Tx is not atomic, not filtering
+			txsToProcess = append(txsToProcess, txs[i])
+			continue
+		} else {
+			// Tx is atomic
+			// Create atomic group
+			atomicGroup := common.AtomicGroup{}
+			for i < len(txs) && txs[i].AtomicGroupID == atomicGroupID {
```
tclemos, line 1185: In order to avoid this inner tx loop and reduce a bit the complexity of this code, what do you think about grouping the txs in a map where the key is the AtomicGroupID and the value is a slice of the tx that compose the AtomicGroup? 

With this, we could loop through the txs once to build the map and then loop through the map once to process all the transactions.

txselector/txselector.go
```diff
@@ -1121,6 +1125,94 @@ func filterFailedAtomicGroups(
...
 }
 
+// filterInvalidAtomicGroups split the txs into the ones that can be porcessed
+// and the ones that can't because they belong to an AtomicGroup that is impossible to forge
+// due to missing or bad ordered txs
+func filterInvalidAtomicGroups(
+	txs []common.PoolL2Tx,
+) (txsToProcess []common.PoolL2Tx, filteredTxs []common.PoolL2Tx) {
+	// validate if an atomic group is valid in terms of requested txs
+	isValid := func(atomicGroup common.AtomicGroup) bool {
```
tclemos, line 1135: I'm not sure about the impact of evaluating this function variable every time this function is called, maybe we could make this as a static private function in the same package with the name isAtomicGroupValid to help to make this filterInvalidAtomicGroups less complex and improve code readability.

txselector/errors.go
tclemos, line 70: typo in the last word `tranaction`

in order to avoid mistakes like this, if you are a `VS Code` user, this extension can help https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

txselector/errors.go
```diff
@@ -65,4 +65,11 @@ const (
...
+	// ErrInvalidAtomicGroupCode error code
+	ErrInvalidAtomicGroupCode int = 17
```
tclemos, line 72: shouldn't this be 18? I mean, the `ErrNoAvailableSlotsCode` is already set to 17, and looks like these codes are part of a sequence.

txselector/errors.go
```diff
@@ -65,4 +65,11 @@ const (
...
+	// ErrInvalidAtomicGroupType error type
+	ErrInvalidAtomicGroupType string = "ErrNoAvailableSlots"
```
tclemos, line 74: shouldn't this value be "InvalidAtomicGroupType", it seems to be a copy and paste mistake

tclemos (changes requested):

txselector/txselector.go
arnaubennassar: This was actually my first attempt, but this was resulting in modifying the original order, which complicated the tests. So instead of fixing the test I did this, shame on me xD

mfcastellani (approved):

tclemos (approved):

PR #1017, opened 2021-07-21
Minimal PoC

### LINKED ISSUE #1020: Coordinator network initial PoC
## Rationale

Create a first iteration that allows two peers to send/receive txs using libp2p pubsub

arnaubennassar:
Closes #1020

### What does this PR does?

Initial proof of concept where to peers are added to the pubsub network, one sends a tx and the other receive it.
Note that the approach is to have something working fast hence there is a lot of hardcoded stuff and related. In terms of review, please take special attention to naming and code architecture. Of course any review is welcome, but it's very likely that I decide to implement suggestions in future PRs, since this is just the PoC!

## Checklist

review them:

- [ ] Include tests


api/coordinatornetwork/coordinatornetwork.go
```diff
@@ -0,0 +1,118 @@
...
+	ctx := context.Background()
+	// create a new libp2p Host that listens on a random TCP port
+	cfgOpts := libp2p.ListenAddrStrings("/ip4/0.0.0.0/tcp/" + port)
+	randNum, err := strconv.Atoi(port)
+	if err != nil {
+		return CoordinatorNetwork{}, err
+	}
+	// TODO: generate ID from coordinator's priv key
+	r := mrand.New(mrand.NewSource(int64(randNum)))
+	priv, _, err := crypto.GenerateKeyPairWithReader(crypto.RSA, 2048, r)
```
jeffprestes, line 52: Maybe here we can use the key already imported into the hermez node.

api/coordinatornetwork/txspool.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+		log.Debug(psTxsPool.self.Pretty(), ": received tx from ", msg.ReceivedFrom.Pretty())
+		tx := new(common.PoolL2Tx)
+		err = json.Unmarshal(msg.Data, tx)
+		if err != nil {
+			continue
+		}
+		log.Info("tx received: ", string(msg.Data))
+		// send valid messages onto the Messages channel
+		psTxsPool.Txs <- tx
```
jeffprestes, line 95: How do think from here the transaction can be included at hermez-node transaction pool?

jeffprestes (approved):
For a PoC, LGTM. 
Congrats for this quick evolution.

api/coordinatornetwork/txspool.go
arnaubennassar: My idea is to separate the necessary logic from the `api` in a way that the same function can be used through the HTTP API and the PubSub network: something like: Interface (HTTP, PubSub, gRPC, ...) ==> handlers (actual logic)

api/coordinatornetwork/coordinatornetwork.go
arnaubennassar: Yeah it would be awesome to find a way to derivate the keys in a way that:

- Private key (BJJ or Ethereum) ==> libp2p private key (deterministic)
- Public key (BJJ or Ethereum) ==> libp2p public key (deterministic)
- Security is preserved

This way the network could allow to seamlessly communicate to specific coordinators through p2p by just reading on-chain data

mfcastellani (approved):

PR #1031, opened 2021-07-23
Avoid rewrite config and service files if already exists

mfcastellani:
### What does this PR does?

do not overwrite current settings if already exists.

### How to test?

run `make install` twice

## Checklist

review them:

- [ ] Include tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -201,8 +201,6 @@ else
 	echo "  - please, check the ./cmd/heznode/cfg.builder.toml!"
 	test -f ./cmd/heznode/cfg.builder.toml
 endif
-	echo "  > Copying hez binary to /usr/local/bin"
-	cp dist/heznode /usr/local/bin/heznode
 ifneq ("$(wildcard /etc/hermez/config.toml)","")
 	echo "  > Config file already exists - ignored."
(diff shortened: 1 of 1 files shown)


jeffprestes (dismissed):

tclemos:
When I read the PR description I thought the approach would be just skipping the config copy or/and maybe displaying a message about it.

Accordingly to the suggested test to run `make install` twice, if you run `make install` more than twice, it will cause the same issue as before this PR:

- **first install**: copy from `cmd/heznode/cfg.builder.toml` to `/etc/hermez/config.toml`
- **second install**: backup `/etc/hermez/config.toml` as `/etc/hermez/config.toml.old` and repeat **first install**
- **third and next installs**: since the **`second install**backed up the cfg as `cfg.builder.toml.old` and the `cfg.builder.toml` is the same as the one contained in the folder `cmd`, this execution will replace the current `old` with a new one 

did you thought about this? is this important?

jeffprestes (dismissed):

marioidival (approved):
LGTM

tclemos (approved):
🎉

PR #1037, opened 2021-07-27
Feature/#711 default config

### LINKED ISSUE #711: Add default values to config params
It would be good to have default values for config params (where it makes sense) and make them optional to keep the file small and clean. 

This would also improve user experience by requiring less changes to config file from the user on node updates, because right now on each addition/removal/name change of config parameter in the code the user is required to update the config file.

ARR552:
Closes #711

### What does this PR does?

This PR adds the functionality to support configuring the hermez node via env variables. If this env variables are not setted, then it uses the variables inside the config file. If there are some config params that are still missing, the hermez node will use the default ones.
Also It uses the config parser as library

Batches forged:
https://explorer.internaltestnet.hermez.io/batch/27288  using env variables
https://explorer.internaltestnet.hermez.io/batch/27301 using config file
https://explorer.internaltestnet.hermez.io/batch/29058  without setting ethscan parameters
#####################
https://explorer.internaltestnet.hermez.io/batch/29177  using env variables without config file
https://explorer.internaltestnet.hermez.io/batch/29180 without etherscan
https://explorer.internaltestnet.hermez.io/batch/29183  with env and config file
https://explorer.internaltestnet.hermez.io/batch/29184  using wrong config file name and env
https://explorer.internaltestnet.hermez.io/batch/29186  using only config file

### How to test?

Run hermez node setting env variables.
Run hermez node without env variables or toml file to check that it is getting the default variables correctly.

## Checklist

review them:

- [ ] Include tests
- [X] Update documentation (*.md) (if needed)


marioidival (dismissed):
LGTM

config/README.md
```diff
@@ -0,0 +1,146 @@
+# Configuration
```
jeffprestes, line 1: I believe all environment variable should have `HEZNODE_` prefix to avoid collision with other applications that might have in the server.

config/default.go
```diff
@@ -0,0 +1,150 @@
...
+URLExtraParams = "&vs_currencies=usd"
+Addresses="0=0x0000000000000000000000000000000000000000,2=0xdac17f958d2ee523a2206206994597c13d831ec7,8=ignore"
+
+[Debug]
+APIAddress = "0.0.0.0:12345"
+MeddlerLogs = true
+GinDebugMode = true
+
+[StateDB]
+Path = "/tmp/iden3-test/hermez/statedb"
```
jeffprestes, line 44: It should use the same values than defined at heznode.service. Please check the `make install`  and with @mfcastellani

jeffprestes (changes requested):

marioidival (approved):

config/default.go
mfcastellani:
```suggestion
Path = "/var/hermez/statedb"
```

config/default.go
```diff
@@ -0,0 +1,150 @@
...
+MinFeeUSD    = 0.0
+MaxFeeUSD    = 50.0
+TTL          = "24h"
+PurgeBatchDelay = 10
+InvalidateBatchDelay = 20
+PurgeBlockDelay = 10
+InvalidateBlockDelay = 20
+
+[Coordinator.TxSelector]
+Path = "/tmp/iden3-test/hermez/txselector"
```
mfcastellani, line 103:
```suggestion
Path = "/var/hermez/txselector"
```

config/default.go
```diff
@@ -0,0 +1,150 @@
...
+
+[Coordinator.BatchBuilder]
+Path = "/tmp/iden3-test/hermez/batchbuilder"
```
mfcastellani, line 106:
```suggestion
Path = "/var/hermez/batchbuilder"
```

config/default.go
```diff
@@ -0,0 +1,150 @@
...
+Attempts = 4
+AttemptsDelay = "500ms"
+TxResendTimeout = "2m"
+NoReuseNonce = false
+MaxGasPrice = 2000
+MinGasPrice = 5
+GasPriceIncPerc = 10
+
+[Coordinator.EthClient.Keystore]
+Path = "/tmp/iden3-test/hermez/ethkeystore"
```
mfcastellani, line 126:
```suggestion
Path = "/var/hermez/ethkeystore"
```

config/default.go
```diff
@@ -0,0 +1,150 @@
...
+Fixed = 600000
+L1UserTx = 15000
+L1CoordTx = 8000
+L2Tx = 250
+
+[Coordinator.API]
+Coordinator = true
+
+[Coordinator.Debug]
+BatchPath = "/tmp/iden3-test/hermez/batchesdebug"
```
mfcastellani, line 139:
```suggestion
BatchPath = "/var/hermez/batchesdebug"
```

mfcastellani (changes requested):

arnaubennassar (changes requested):
Maybe I'm missing something, but since there are lots of fields in the structs with the `required` tag, parsing this object will still have the same problems and errors that we have today. As of my understanding we should stop using this tags, since we will always provide a default value

jeffprestes (approved):

jeffprestes:
> Maybe I'm missing something, but since there are lots of fields in the structs with the `required` tag, parsing this object will still have the same problems and errors that we have today. As of my understanding we should stop using this tags, since we will always provide a default value

Yeah. It may become optional instead of required however, as you've said, as we have default values it should work as the required values might be obtained from default values. Did you find any issue starting a server?

arnaubennassar:
A bit out of the scope for this PR review, but wanted to mention a few things about github.com/hermeznetwork/config-library:

- By context it's obvious that this is a library (package and library are quite equivalent words), any Go package that can be imported could be named `xxx-library`, therefore it's a bit redundant. 
- By using the `-` in `config-library`, most people will create an alias on the import as you did here: `configLibrary "github.com/hermeznetwork/config-library"`, in general it's a good idea to avoid characters like `-` in package names

I'm sharing this as tips/tricks that I consider when I've to do naming related stuff, which as you know it's one of the most difficult parts of our job.

ARR552:
> Maybe I'm missing something, but since there are lots of fields in the structs with the `required` tag, parsing this object will still have the same problems and errors that we have today. As of my understanding we should stop using this tags, since we will always provide a default value

Those params are still required by hermez-node. However, if the user doesn't set any value for a specific param, the default value will be used by hermez node.

arnaubennassar:
> > Maybe I'm missing something, but since there are lots of fields in the structs with the `required` tag, parsing this object will still have the same problems and errors that we have today. As of my understanding we should stop using this tags, since we will always provide a default value
> 
> Yeah. It may become optional instead of required however, as you've said, as we have default values it should work as the required values might be obtained from default values. Did you find any issue starting a server?

Haven't tried this branch, but I thought that the main idea was to be able to use a smaller config or a few ENV vars with just the specific stuff that wants to be configured in particular. But by looking at the code, it seems that we still need the full config file to start the node.

Example:

- We add a new var into the config file, which have a rational default value, although following the current logic we mark it as `required`
- When people update to this changes they will need to go through the config file and manually add the new value, otherwise when trying to parse the config file there will be an error and the node won't start

Note that the example has happened to me many time by changes on the config file across versions

arnaubennassar:
> > Maybe I'm missing something, but since there are lots of fields in the structs with the `required` tag, parsing this object will still have the same problems and errors that we have today. As of my understanding we should stop using this tags, since we will always provide a default value
> 
> Those params are still required by hermez-node. However, if the user doesn't set any value for a specific param, the default value will be used by hermez node.

But the config file is still required to start the node right? So if I don't have ALL the (required) values on the config file, the node will not start.

IMO (not saying that this has to be implemented by this PR) config file shouldn't even be required to start the node, it's perfectly fine to just use ENV vars. Of course we need logic to check if some values are inputed by the user or not, because not everything can work with default values

jeffprestes:
> > > Maybe I'm missing something, but since there are lots of fields in the structs with the `required` tag, parsing this object will still have the same problems and errors that we have today. As of my understanding we should stop using this tags, since we will always provide a default value
> > 
> > 
> > Yeah. It may become optional instead of required however, as you've said, as we have default values it should work as the required values might be obtained from default values. Did you find any issue starting a server?
> 
> Haven't tried this branch, but I thought that the main idea was to be able to use a smaller config or a few ENV vars with just the specific stuff that wants to be configured in particular. But by looking at the code, it seems that we still need the full config file to start the node.
> 
> Example:
> 
> * We add a new var into the config file, which have a rational default value, although following the current logic we mark it as `required`
> * When people update to this changes they will need to go through the config file and manually add the new value, otherwise when trying to parse the config file there will be an error and the node won't start
> 
> Note that the example has happened to me many time by changes on the config file across versions

The idea is you don't need a config file anymore. @ARR552 have you tried to start a hermez node without any config file nor env variables?

ARR552:
> > > > Maybe I'm missing something, but since there are lots of fields in the structs with the `required` tag, parsing this object will still have the same problems and errors that we have today. As of my understanding we should stop using this tags, since we will always provide a default value
> > > 
> > > 
> > > Yeah. It may become optional instead of required however, as you've said, as we have default values it should work as the required values might be obtained from default values. Did you find any issue starting a server?
> > 
> > 
> > Haven't tried this branch, but I thought that the main idea was to be able to use a smaller config or a few ENV vars with just the specific stuff that wants to be configured in particular. But by looking at the code, it seems that we still need the full config file to start the node.
> > Example:
> > 
> > * We add a new var into the config file, which have a rational default value, although following the current logic we mark it as `required`
> > * When people update to this changes they will need to go through the config file and manually add the new value, otherwise when trying to parse the config file there will be an error and the node won't start
> > 
> > Note that the example has happened to me many time by changes on the config file across versions
> 
> The idea is you don't need a config file anymore. @ARR552 have you tried to start a hermez node without any config file nor env variables?

It is not possible without any configuration. At least, the user has to set and load its ethereum account, bjj and some other params. Also, it would be great if we can define a proper default config file or maybe you can send me the config file that we are using in mainnet. Do you have this info?

mfcastellani (approved):

arnaubennassar (dismissed):

arnaubennassar (approved):

mfcastellani (approved):

arnaubennassar (approved):
Not asking for it in this PR, but improvement idea: use a tag in the struct for `default`, so the config library can set the default value if it's not provided anywhere else

tclemos (approved):

PR #1041, opened 2021-07-27
Coordnet/wan discovery

### LINKED ISSUE #1021: Coordinators network: enable discovery using on chain data
## Rationale

We should take advantage of the things we know from the chain about other coordinators in order to onboard new comers to the PubSub network. Since coordinators have he ability to store a URL on the SC when they register we could use URL:well known port to find them.

arnaubennassar:
Closes #1021 

### What does this PR does?

Setup WAN discovery for coordinators network. Note that right now tests are done by manually using the ENV var `ADDR`, however it only uses this address and successfully shares one published transaction among many coordinators (tested with two subscribed servers and publishing from my laptop). 

This proofs that we have enough with on chain data to bootstrap the network.

### How to test?

With at least two machines:

Machine A: `FAKE_COORDNET=yes go test ./api/coordinatornetwork/... --run TestPubSubFakeServer -count=1 -v`. Wait until you see the log `libp2p ID: Qm...`
Machine B: `PUBLISH=yes FAKE_COORDNET=yes ADDR="/dns/<URL>/tcp/3598/p2p/<libp2p ID>" go test ./api/coordinatornetwork/... --run TestPubSubFakeServer -count=1 -v -count=1`. Replace `<URL>` for the URL of Machine A and `<libp2p ID>` for the log from machine A (`Qm...`)

## Checklist

review them:

- [ ] Include tests


api/coordinatornetwork/libp2p.go
```diff
@@ -0,0 +1,212 @@
...
+	dht "github.com/libp2p/go-libp2p-kad-dht"
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+	yamux "github.com/libp2p/go-libp2p-yamux"
+	"github.com/libp2p/go-tcp-transport"
+	"github.com/mr-tron/base58"
+	"github.com/multiformats/go-multiaddr"
+	"github.com/multiformats/go-multihash"
+)
+
+func setupHost(ctx context.Context, registeredCoordinators []common.Coordinator) (host.Host, *dht.IpfsDHT, error) {
```
jeffprestes, line 29: I am wondering where you plan to setup and register the coordinators slice. At cmd/main.go? 
Also, you need to be aware this list might be upgradable without need to restart hermeznode .  Maybe once in hour check if there is a new coordinator.

jeffprestes (approved):

api/coordinatornetwork/libp2p.go
arnaubennassar: > I am wondering where you plan to setup and register the coordinators slice. At cmd/main.go?

I believe this could end up being used outside of `hermez-node`, therefore I prefer to have a flexible approach in the sense of moving this logic into `node/node.go`. In our case I'm considering a few options:

a) Get the list from the DB, this is fast and easy but it has a problem: nodes that have just started for the first tie won't have coordinators stored on the DB (they have not synched anything yet)
b) Read the registered coordinators directly from the SC
c) Mix of the previous (if there are less than N coordinators on the DB, check it on Ethereum)

> Also, you need to be aware this list might be upgradable without need to restart hermeznode

Absolutely, the function `func (coordnet CoordinatorNetwork) FindMorePeers() error` should do the trick, although is hard to test this, if we find that this is not enough we could use the coordinator list from DB/SC again. In my experience while testing the connections were being handled nicely by using this function (nodes subscribed to the pubsub were appearing/disappearing correctly on each machine)

jeffprestes: > b) Read the registered coordinators directly from the SC
I believe this is most reliable option for starting.

>there is no need for all the nodes to be connected to all the nodes, from the official documentation
It makes sense. In general Ethereum nodes are connected to 5 other nodes. Six is a good number.

jeffprestes: > b) Read the registered coordinators directly from the SC
I believe this is most reliable option for starting.

>there is no need for all the nodes to be connected to all the nodes, from the official documentation
It makes sense. In general Ethereum nodes are connected to 5 other nodes. Six is a good number.

mfcastellani (approved):

Issue #1030, opened 2021-07-22
Add public marshalers and unmarshalers for API requests and responses  


arnaubennassar:
## Rationale

Right now anyone trying to integrate with the node in Go has to built custom logic to parse request/response bodies from/to the API (including the SDK), this forces client to have some sort of duplicated structs, instead of using the ones defined in this repo.

This is specially relevant for the structs used in POST methods requests. Although they are already implemented (`common.PoolL2Tx`, `common.AtomicGroup`) except for `common.CreateAccountAuth`.

Note that there are lot's of such structs, and probably it will be overkill to implement this in a single PR

jeffprestes:
The issue (related to https://github.com/hermeznetwork/hermez-go-sdk/issues/13) in Hermez Go SDK needs that this issue has been completed to be fixed.

jeffprestes:
All MarshalJSON needs to return a public common struct instead of interface{}

mfcastellani:
@jeffprestes and @arnaubennassar after some time looking at the structures present in the common package, I would like to discuss with you a way to standardize and improve the description of some of the structures in the package.

I'll start with PoolL2Tx. If this is something that will be used outside the project, that will be exposed, the name is very bad. It is not a pool, but a single transaction. Also, I don't see another type of transaction being externalized in our common package, so this structure could simply be called a transaction. This should extend to our api, so we would have a GET transaction endpoint that would receive an id and take care of fetching it inside the system, no matter where it is.

Another point is the way things are organized in files. We have for example Nonce, which is defined inside account. And it is used within pooll2tx.go. If Nonce is a reusable data type I believe it should have its own file inside common.

Another thing that caught my attention is that there are several "builders" such as NonceFromBytes or AccountFromBigInts that are not linked to any structure. Taking an example from Geth, if you want to call the Dial method to create a client you do ethclient.Dial(URL). In our case everything is within common. I don't know if it would be over-precious to have something like common.account.Account or common.account.FromBigInts, which in my opinion would be more descriptive and simpler to use.

I am raising these points since the objective of this PR is to have a better usability of the structures within the common, as said in this PR and related ones. And just bringing the json serialize and deserialize methods over here isn't going to make this API any prettier.

I would like your opinion. In time, I have already started to make some of these changes.

jeffprestes:
[...]

PR #1042, opened 2021-07-27
Add common.nonce package

mfcastellani:
### What does this PR does?

Extract `nonce` structure and related methods to an isolated package in 
order to provide to external developers a better experience

This code is a implementation related to this comment, https://github.com/hermeznetwork/hermez-node/issues/1030#issuecomment-887817946, to better explain the whole idea. 

### How to test?

`make test`

## Checklist

review them:

- [ ] Include tests


ARR552 (dismissed):

arnaubennassar (changes requested):
Like the idea to separate the nonce, but IMO it's overkill to do it in a separated package. I'd rather do it in `common/nonce.go`. 

The reason for this opinion is consistency: nonce is just one of the many many types that life under `common`, if we do this change I believe we should treat all the other types in the same way (one type = one package), which will result in tons of directories having most of the times just one file.

I'm not saying that this is a bad idea (although I don't like it), but I believe that if we do this change we should do the same with all the other types, which will have a huge impact.

I've read your comment in #1030 and I fully agree on all what you said about the nonce, however I don't see how this is connected to having it in a sub-package, so please could you:

a) Consider moving `common/nonce/nonce.go` ==> `common/nonce.go`
b) Give a bit of context/rationale about the decision of having a new layer of sub-packages

common/nonce/errors.go
```diff
@@ -0,0 +1,6 @@
+package nonce
```
jeffprestes, line 1: It's nice to bring nonce items from account to common package. It's going to be very useful.
However, I don't think it needs a entire package for it. I believe the items in common package should stay on it. We will avoid 
`common.xyz.xyz` . As I mentioned before I think w should get GETH as reference.

jeffprestes (changes requested):

common/nonce/errors.go
mfcastellani: @jeffprestes and @arnaubennassar, explaining better the idea:

The main idea here is to actually have the types isolated in highly specialized packages so that common users know exactly where to find what.

Since everything is in common there is no specialization of the package. It's like a big disorganized toolbox where we find hammer, screwdrivers, and that kitchen knife we ​​once lost.

I understand that dividing just by dividing is foolishness and an excess of preciosity, but in this case, we are giving the end user, who does not have all the knowledge we have about the platform, a hierarchical toolbox, where he can easily find what he is looking for.

This approach **is the same as used in the standard go library**, which does not have a common package. Everything there is inside packages. And there you don't have `common.NonceFromBytes`. You have `nonce.FromBytes`, such as `ascii85.Decode` or `hex.Decode`, not `encoding.AscII85Decode` or `encoding.HexDecode` (note that nonce in this case is not a variable but the package).

This approach makes it much easier for a newcomer to our framework to use our common package. And, in the model I proposed, the occurrence of `nonce.Nonce` exists only in the type declaration. So, the main idea here is have, for a hypothetical structure Potato  the types, related errors, marshallers, etc, in a package called potato with mostly of the methods attached to the Potato struct, but when is not possible (like create a Potato from another struct, for example, PotatoFromTomato - instead having common.PotatoFromTomato we can have potato.FromTomato wich is much more declarative). 

Let me know your thoughts. But again, I ask you to look at this implementation not as hermez developers with a skewed view, but as an outsider who will use the common package.

arnaubennassar: I got your point, and there is some cases that falls into what you're mentioning such as:

```go
package common

type PoolL2TxState string

const (
	// PoolL2TxStatePending represents a valid L2Tx that hasn't started the
	// forging process
	PoolL2TxStatePending PoolL2TxState = "pend"
	// PoolL2TxStateForging represents a valid L2Tx that has started the
	// forging process
	PoolL2TxStateForging PoolL2TxState = "fing"
	// PoolL2TxStateForged represents a L2Tx that has already been forged
	PoolL2TxStateForged PoolL2TxState = "fged"
	// PoolL2TxStateInvalid represents a L2Tx that has been invalidated
	PoolL2TxStateInvalid PoolL2TxState = "invl"
)
```

Certainly this would look better:

```go
package pooll2tx

type State string

const (
	// StatePending represents a valid L2Tx that hasn't started the
	// forging process
	StatePending PoolL2TxState = "pend"
	// ...
)
```

However moving all the types into sub-packages feels too much for me. Maybe we can find a balanced solution where we a few sub-packages that make sense among them? 

Also note that if we follow your approach we have to be very careful about cyclic dependencies

jeffprestes: I agree with @arnaubennassar . We need to have a balance. Maybe for items related to important business items a specific package makes sense but for accessory items I believe a toolbox is not a problem.

jeffprestes (approved):

arnaubennassar (dismissed):

jeffprestes (approved):

jeffprestes (approved):

tclemos (approved):

PR #1049, opened 2021-07-28
Remove use of interface

mfcastellani:
Related to https://github.com/hermeznetwork/hermez-node/issues/1030 and https://github.com/hermeznetwork/hermez-go-sdk/issues/13

### What does this PR does?

Avoid use of interface for `PoolTxAPI` structure, method `MarshalJSON`

### How to test?

`make test`

## Checklist

review them:

- [ ] Include tests


db/l2db/views.go
```diff
@@ -38,64 +38,105 @@ type PoolTxAPI struct {
...
-			"id":               tx.TokenID,
-			"itemId":           tx.TokenItemID,
-			"ethereumBlockNum": tx.TokenEthBlockNum,
-			"ethereumAddress":  tx.TokenEthAddr,
-			"name":             tx.TokenName,
-			"symbol":           tx.TokenSymbol,
-			"decimals":         tx.TokenDecimals,
-			"USD":              tx.TokenUSD,
-			"fiatUpdate":       tx.TokenUSDUpdate,
+	type jsonToken struct {
```
jeffprestes, line 58: The is struct is perfect. However, it needs to be defined in common package and public in order to be used for external lib/binaries such as SDK. Could you move both to other package or at least define them as public in this package?

jeffprestes (changes requested):

db/l2db/views.go
arnaubennassar: If we're moving this into common, I'd like to suggest to do something like:

```go
// common/pooll2tx.go
package common

type PoolL2TxAPI struct // similar to l2db.PoolL2TxAPI BUT with neasted token struct instead of flat
func (tx PoolTxAPI) MarshalJSON() ([]byte, error)
func (tx *PoolTxAPI) UnmarshalJSON(data []byte) error


// db/l2db/views.go
package l2db

type poolTxAPIView struct // same as l2db.PoolTxAPI
func poolTxViewToAPI(tx poolTxAPIView) common.PoolL2TxAPI
``` 

Rationale: we're using a flat struct in order to be able to put the data from the SQL rows, and we were making the nesting of the sub-struct (token) at the JSON marshal level, however if this struct has to be used elsewhere, it's very ugly for it to  be flat. Therefore I think the best option is to have the good non-flat struct at `common`, and let the `l2db` deal with the limitations of the underlying DB (note that this is very specific to SQL, if one day we switch to a noSQL it's likely that this won't be a problem and we can return the struct with extra transformation).

arnaubennassar (changes requested):

db/l2db/views.go
marioidival: I think this struct should not be moved to other place. Also, already exist a similar struct there https://github.com/hermeznetwork/hermez-node/blob/f39996efbb9130749bc6476bd7fd257b9e20af10/db/historydb/views.go#L244. 

If we're think to do a move (refactor), we should do a really big effort to this, because on the project will there repeated struct's. I think the purpose of PR is done.

marioidival (dismissed):

db/l2db/views.go
jeffprestes: I guess there is a misunderstanding here. The issue was initially opened to allow SDK and other libraries to reuse structs  returned in API responses. and avoid we need to rebuild the same struct in other repo. It was not intent a full refactor. That's why without make them public the PR does not achieve issue's requirement.

jeffprestes (dismissed):

common/pooll2txapi.go
```diff
@@ -0,0 +1,147 @@
...
+		ItemID               uint64                 `json:"itemId"`
+		TxID                 TxID                   `json:"id"`
+		Type                 TxType                 `json:"type"`
+		FromIdx              Idx                    `json:"fromAccountIndex"`
+		EffectiveFromEthAddr *ethCommon.Address     `json:"fromHezEthereumAddress"`
+		EffectiveFromBJJ     *babyjub.PublicKeyComp `json:"fromBJJ"`
+		ToIdx                *Idx                   `json:"toAccountIndex"`
+		EffectiveToEthAddr   *ethCommon.Address     `json:"toHezEthereumAddress"`
+		EffectiveToBJJ       *babyjub.PublicKeyComp `json:"toBJJ"`
+		Amount               big.Int                `json:"amount"`
```
arnaubennassar, line 77: This fields are not following the API format

common/pooll2txapi.go
```diff
@@ -0,0 +1,147 @@
...
+		MaxNumBatch          uint32                 `json:"maxNumBatch"`
+		Info                 *string                `json:"info"`
+		ErrorCode            *int                   `json:"errorCode"`
+		ErrorType            *string                `json:"errorType"`
+		Signature            babyjub.SignatureComp  `json:"signature"`
+		Timestamp            time.Time              `json:"timestamp"`
+		RqFromIdx            *Idx                   `json:"requestFromAccountIndex"`
+		RqToIdx              *Idx                   `json:"requestToAccountIndex"`
+		RqToEthAddr          *ethCommon.Address     `json:"requestToHezEthereumAddress"`
+		RqToBJJ              *babyjub.PublicKeyComp `json:"requestToBJJ"`
```
arnaubennassar, line 90: This fields are not following the API format

common/pooll2txapi.go
```diff
@@ -0,0 +1,147 @@
...
+		RqTokenID            *TokenID               `json:"requestTokenId"`
+		RqAmount             big.Int                `json:"requestAmount"`
```
arnaubennassar, line 92: This fields are not following the API format

arnaubennassar (changes requested):
Apart from the comments, I'd like to only have the JSON marshaling implemented only once, either on l2db or common. Also try to use the type from common in the API, it will help you to ensure that the marshaling is working well

common/pooll2txapi.go
mfcastellani: Hello. Sorry, i did not understood this comment and others 👇 . As you can see here https://github.com/hermeznetwork/hermez-node/pull/1049/commits/20f8497c0118523c4f77a1d32d4c9fe8bd070ee7#diff-d529406a2271fd28bec09e70ede2455ba1b484a87d25250e37f3f6390dfc143dL60 I just copied the JSON field name as is, no changes.

mfcastellani:
> Apart from the comments, I'd like to only have the JSON marshaling implemented only once, either on l2db or common. Also try to use the type from common in the API, it will help you to ensure that the marshaling is working well

@arnaubennassar change the types to use the common types will need a big refactor in the database layer. If you agree, since @jeffprestes needs this asap, I want to merge this PR after solve my doubts regarding the `This fields are not following the API format` comments and after that start another one with this refactor, ok?

arnaubennassar:
>  I want to merge this PR after solve my doubts regarding the `This fields are not following the API format` comments and after that start another one with this refactor, ok?

Makes sense. About the `This fields are not following the API format`, let's have a look into some specific fields, focusing in the requirements for the SDK:

```go
// common/pooll2txapi.go

type PoolL2TxAPI struct {
        // ...
        FromIdx Idx // Will fail to unmarshal, because it doesn't have the json tag, also because received JSON will look like "hez:ETH:3498" instead of 3498
        EffectiveToEthAddr *ethCommon.Address // Will fail to unmarshal, because it doesn't have the json tag, also because received JSON will look like "hez:0xA78cD..." instead of "0xA78cD..."
}

func (tx *PoolL2TxAPI) UnmarshalJSON(data []byte) error {
	err := json.Unmarshal(data, tx)
	if err != nil {
		return err
	}
	return nil
}
```

This covers the Unmarshaling problems, for the Marshaling the json tags are right, but the types have the same problem ("hez:ETH:3498" vs 3498, "hez:0xA78cD..." vs "0xA78cD..."). Note that this are examples for Idx and EthAddr but BJJ and any big.Int have similar situations where the obvious way of marshaling is not used and we have "custom" way for JSON representation

arnaubennassar (dismissed):
Approving because I understand this is blocking the SDK. But there's a few comments I'd like to make:

Opinionated stuff:

- `l2db.PoolTxAPIView` should be private type, and functions from `l2db` package should return `apitypes.PoolTxAPI` instead (by calling `ToAPI()` before returning)
- `l2db.PoolTxAPIView` shouldn't implement marshal/unmarshal
- `api/txspool.go` should use `apitypes.PoolTxAPI` instead of `l2db.PoolTxAPI`

IMO important:

- `api/txspool_test.go` should use `apitypes.PoolTxAPI` instead of `testPoolTxReceive`
- `apitypes.PoolTxAPI` marshal/unmarshal should be tested. I suggest implementing the previous points because by doing so the marshal/unmarshal will be fully tested including validation against swagger specification

Improvement ideas

- `apitypes.PoolTxAPI` could use "common" types instead of "apitypes" (example `FromIdx` could be type Idx instead of `HezIdx`). IMO SDK and anything else should work with the "native" types, to string representations. Imagine you're using the API/SDK you will want to do stuff like check if you're the sender of this tx by comparing your idx vs the tx FromIdx. In this case you would be comparing `734 == "hez:ETH:734"`... I thing that this sort of representation should exist only at the Marshal/Unmarshal level. Please have a look on how `common.PoolL2Tx` deals with it [here](https://github.com/hermeznetwork/hermez-node/blob/develop/common/pooll2tx.go#L452)
- I don't like the name `PoolTxAPI` (and probably it was set by me in the past xD) it will be confusing for users to have `PoolL2Tx` and `PoolTxAPI`, the difference between this structs is that the API type doesn't have functions to sign, generate hash, set type, ... Instead it has extended info such as full Token information

common/apitypes/pooll2txapi.go
```diff
@@ -0,0 +1,147 @@
...
+	"encoding/json"
+	"time"
+
+	ethCommon "github.com/ethereum/go-ethereum/common"
+	"github.com/hermeznetwork/hermez-node/common"
+	"github.com/iden3/go-iden3-crypto/babyjub"
+)
+
+// PoolL2TxAPI represents a L2 Tx pool with extra metadata used by the API
+type PoolL2TxAPI struct {
```
jeffprestes, line 13: Unfortunately, the `json` is missing here. The SDK and other 3rd-party softwares would need the json tags in here to Unmarshall when it receives the return the json string of the API REST call. Could you please add them here?

db/l2db/views.go
```diff
@@ -5,13 +5,13 @@ import (
...
-	"github.com/hermeznetwork/hermez-node/api/apitypes"
 	"github.com/hermeznetwork/hermez-node/common"
+	"github.com/hermeznetwork/hermez-node/common/apitypes"
 	"github.com/iden3/go-iden3-crypto/babyjub"
 )
 
-// PoolTxAPI represents a L2 Tx pool with extra metadata used by the API
-type PoolTxAPI struct {
+// PoolTxAPIView represents a L2 Tx pool with extra metadata used by the API
+type PoolTxAPIView struct {
```
jeffprestes, line 14: If this struct is also used in API return I believe it also needs the json tag

jeffprestes (changes requested):

jeffprestes:
> * `l2db.PoolTxAPIView` should be private type, and functions from `l2db` package should return `apitypes.PoolTxAPI` instead (by calling `ToAPI()` before returning)

I agree. Because the 3rd party libraries would need the struct in order to consume the API.


> api/txspool_test.go should use apitypes.PoolTxAPI instead of testPoolTxReceive

Agreed.

> apitypes.PoolTxAPI marshal/unmarshal should be tested. I suggest implementing the previous points because by doing so the marshal/unmarshal will be fully tested including validation against swagger specification

These tests are very important because this is the main goal of this issue/PR. An external APP should be able to use these structs to unmarshall/marshall API calls.

> apitypes.PoolTxAPI could use "common" types instead of "apitypes" (example FromIdx could be type Idx instead of HezIdx). IMO SDK and anything else should work with the "native" types, to string representations. Imagine you're using the API/SDK you will want to do stuff like check if you're the sender of this tx by comparing your idx vs the tx FromIdx. In this case you would be comparing 734 == "hez:ETH:734"... I thing that this sort of representation should exist only at the Marshal/Unmarshal level. Please have a look on how common.PoolL2Tx deals with it here

I agree too.


> I don't like the name PoolTxAPI (and probably it was set by me in the past xD) it will be confusing for users to have PoolL2Tx and PoolTxAPI, the difference between this structs is that the API type doesn't have functions to sign, generate hash, set type, ... Instead it has extended info such as full Token information

Let's take the opportunity of this reflection to improve not only the items locations but also the names.

mfcastellani:
> Approving because I understand this is blocking the SDK. But there's a few comments I'd like to make:
> 
> Opinionated stuff:
> 
> * `l2db.PoolTxAPIView` should be private type, and functions from `l2db` package should return `apitypes.PoolTxAPI` instead (by calling `ToAPI()` before returning)
> * `l2db.PoolTxAPIView` shouldn't implement marshal/unmarshal
> * `api/txspool.go` should use `apitypes.PoolTxAPI` instead of `l2db.PoolTxAPI`
> 
> IMO important:
> 
> * `api/txspool_test.go` should use `apitypes.PoolTxAPI` instead of `testPoolTxReceive`
> * `apitypes.PoolTxAPI` marshal/unmarshal should be tested. I suggest implementing the previous points because by doing so the marshal/unmarshal will be fully tested including validation against swagger specification
> 
> Improvement ideas
> 
> * `apitypes.PoolTxAPI` could use "common" types instead of "apitypes" (example `FromIdx` could be type Idx instead of `HezIdx`). IMO SDK and anything else should work with the "native" types, to string representations. Imagine you're using the API/SDK you will want to do stuff like check if you're the sender of this tx by comparing your idx vs the tx FromIdx. In this case you would be comparing `734 == "hez:ETH:734"`... I thing that this sort of representation should exist only at the Marshal/Unmarshal level. Please have a look on how `common.PoolL2Tx` deals with it [here](https://github.com/hermeznetwork/hermez-node/blob/develop/common/pooll2tx.go#L452)
> * I don't like the name `PoolTxAPI` (and probably it was set by me in the past xD) it will be confusing for users to have `PoolL2Tx` and `PoolTxAPI`, the difference between this structs is that the API type doesn't have functions to sign, generate hash, set type, ... Instead it has extended info such as full Token information

Some refactors will be handled in a separate issue - https://github.com/hermeznetwork/hermez-node/issues/1064

jeffprestes (approved):

Issue #1029, opened 2021-07-22
Add unit and integration like tests for coordinators network


arnaubennassar:
## Rationale

It's very complicated to have good unit tests for p2p networks, as the discovery mechanisms that will be used in a real environment will be very different from the ones used in a real environment. Furthermore the used port should be a constant for all coordinators, as the network will be bootstrapped using the URL of the registered coordinators in the SCs + a hardcoded port.

## Implementation

a) use Docker (which is already a recommended test dependency) and create a minimal image that acts as a coordinator but just with the coordinator network feature, this way we can use the same port on multiple instances? 
b) define the port as private variable of the package, so the test can modify it

arnaubennassar:
In order to test how the addresses are fetched from Ethereum (URL  public key) ==> libp2p address, it would be very nice to use geth bakends package to simulate the registration of coordinators that we know it's private key so we can ensure that the libp2p ID is properly generated.

PR #1060, opened 2021-08-03
Coordnet/get coord p2p addr from eth

### LINKED ISSUE #1021: Coordinators network: enable discovery using on chain data

### LINKED ISSUE #1026: Derivate libp2p private key from coordinator key
## Rationale

Use either the Ethereum or BJJ private key to generate the key used for libp2p. This way coordinators don't need to care about another private key. 

Also it would be awesome if public key can also be derivated the same way. By doing this we could enable more advanced communication where anyone can reach a particular coordinator by just knowing some on-chain data

arnaubennassar:
Closes #1026 
Closes #1021 

### What does this PR does?

- Read the list of registered coordinators from Ethreum in order to obtain the URL and find the public key of the coordinator (by fetching the tx that originated the event)
- Generate the libp2p ID from the public key of the coordinators

### How to test?

The added tests have to be run as explained in `eth/README.md`, better tests will be included when facing #1029 however I've run it against mainnet and goerli, and looks like all the coordinators addresses are generated properly  

## Checklist

review them:

- [x] Include tests

### DIFF
--- a/eth/auction.go
+++ b/eth/auction.go
@@ -16,7 +16,6 @@ import (
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	ethCommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
-	"github.com/ethereum/go-ethereum/crypto"
 	ethCrypto "github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/hermeznetwork/hermez-node/common"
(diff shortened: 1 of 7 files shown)


eth/auction.go
```diff
@@ -990,3 +996,135 @@ func (c *AuctionClient) AuctionEventsByBlock(blockNum int64,
...
+// that has been registered so far
+func (c AuctionClient) GetCoordinatorsLibP2PAddrs() ([]multiaddr.Multiaddr, error) {
+	// Get events
+	query := ethereum.FilterQuery{
+		Addresses: []ethCommon.Address{
+			c.address,
+		},
+		Topics: [][]ethCommon.Hash{{logAuctionSetCoordinator}},
+	}
+	logs, err := c.client.client.FilterLogs(context.TODO(), query)
```
jeffprestes, line 1010: I believe `client.client` looks like a code smell. I'd recommend to use names that specify what the client is for.

jeffprestes (changes requested):

eth/auction.go
arnaubennassar: I don't like this either, but this comes from the underlying struct `AuctionClient` that has +30 methods. Changing this would require a considerable amount of changes.

What do you think about opening an issue to address this, and move on with this PR?

Also, I believe that the other SC related structs follow the same pattern, so we could take advantage of a dedicated issue and improve all of them at once

mfcastellani: Issue created for handle this: https://github.com/hermeznetwork/hermez-node/issues/1063

jeffprestes: So, it's event worst because the call is: client.client.client

Please, open an issue asking to change the client variable name showing what's this client for. Whether it's for Ethereum it needs to have AuctionEthClient for instance. Please try to be very specific in the issue's text to avoid misunderstand. Thanks.

mfcastellani (approved):

eth/auction.go
```diff
@@ -990,3 +996,135 @@ func (c *AuctionClient) AuctionEventsByBlock(blockNum int64,
 	}
 	return &auctionEvents, nil
 }
+
+// GetCoordinatorsLibP2PAddrs return the libp2p addr associated to each coordinator
+// that has been registered so far
+func (c AuctionClient) GetCoordinatorsLibP2PAddrs() ([]multiaddr.Multiaddr, error) {
```
jeffprestes, line 1002: Here I think you can be more clear what this client is about. I'd suggest: ```func (ethClient AuctionEthClient) GetCoordinatorsLibP2PAddrs() ([]multiaddr.Multiaddr, error) {```

jeffprestes (changes requested):

ARR552 (approved):

Mikelle (approved):

PR #1072, opened 2021-08-10
Feature/tx selector sort txs by profitability

### LINKED ISSUE #266: In StateDB.ProcessTxs merge calculation of `s.accumulatedFees` and `collectedFees`
This is a low priority optimization.

Currently in ProcessTxs the same amount for coordinator fee is calculated twice, once in s.accumulatedFees (indexed by idx), another for collectedFees (indexed by TokenID).  This could be reduced by only calculating the amount once in s.accumulatedFee, and before returning creating the collectedFees map like this:
```go
collectedFees := make(map[common.TokenID]*big.Int)
coordIdxsMap := s.getTokenIDsFromIdxs(coordIdxs) // update getTokenIDsFromIdxs to return map[common.Idx]common.TokenID
for idx, amount := range s.accumulatedFees {
    collectedFees[coordIdxsMap[idx]] = amonut
}
```

tclemos:
Closes #266.

### What does this PR does?
Adjust the transaction selector to select transactions by profitability instead of only by Nonce.

### How to test?

Send transactions with different sender accounts and fees, the batch forged with these transactions must have the transactions with greater fees selected before the ones with smaller fees between transactions with different sender accounts, but respecting the Nonce sequence for each specific sender account.

These are the batches containing 3000 transactions already forged using the version of this PR to sort the transactions by profitability in the internal testnet network.
image

## Checklist

review them:

- [x] Include tests


jeffprestes (approved):
The code seems to be ok and I am going to approve it. 
However, I just would like to do a small confirmation: if I send a transaction with a wrong nonce the heznode with this modification will block and mark this transaction with a wrong nonce as invalid?

tclemos:
> The code seems to be ok and I am going to approve it.
> However, I just would like to do a small confirmation: if I send a transaction with a wrong nonce the heznode with this modification will block and mark this transaction with a wrong nonce as invalid?

Yes, nothing was changed regarding the rules to decide if a transaction is valid or not, the change just affects the order they are selected.

mfcastellani (approved):

arnaubennassar (approved):

ARR552 (approved):

Mikelle (approved):

PR #1075, opened 2021-08-12
added possibility to update tx with invl status

Mikelle:
### What does this PR does?

Added possibility to update tx with invl status

## Checklist

review them:

- [ ] Include tests

### DIFF
--- a/api/txspool.go
+++ b/api/txspool.go
@@ -94,7 +94,7 @@ func (a *API) putPoolTxByIdxAndNonce(c *gin.Context) {
 	}
 	if receivedTx.State != common.PoolL2TxStatePending || receivedTx.FromIdx != idx || receivedTx.Nonce != nonce {
 		retBadReq(&apiError{
-			Err:  errors.New("tx state is not pend or fromIdx or nonce in request body not equal request uri params"),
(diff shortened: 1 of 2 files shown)


jeffprestes (dismissed):

jeffprestes (dismissed):

api/txspool.go
```diff
@@ -92,9 +92,10 @@ func (a *API) putPoolTxByIdxAndNonce(c *gin.Context) {
 		}, c)
 		return
 	}
-	if receivedTx.State != common.PoolL2TxStatePending || receivedTx.FromIdx != idx || receivedTx.Nonce != nonce {
+	if (receivedTx.State != common.PoolL2TxStatePending && receivedTx.State != common.PoolL2TxStateInvalid) ||
```
tclemos, line 95: I'm not sure about this change.

We want to allow txs with status `invl` to be updated in order to fix the txs, but this change means we want txs to have its `state` updated to `pend` or `invl`, because this is checking the data provided to the request, not the data already in the database, and this data will be used to update the fields in the database if the update succeeds.

In my understanding, this `if` statement shouldn't be changed and any tx update should lead us to a `tx` with a `pend` state in order to make the node process it in the next forge interaction.

Does it make sense?

jeffprestes (approved):

tclemos (approved):

PR #1113, opened 2021-08-24
Improve Sync process to handle blocks with SC Events

tclemos:
### What does this PR does?

Replaces the selection of the next block to sync by a logic that will look for the next block containing events in the SC instead of appending +1.

### How to test?

start a brand new node and sync it


Mikelle (approved):

marioidival (approved):
LGTM

mfcastellani (approved):

synchronizer/synchronizer_test.go
```diff
@@ -1,878 +0,0 @@
-package synchronizer
```
ARR552, line 1: Why delete this test instead of adapt it??

PR #1123, opened 2021-09-02
fix UpdateTxsInfo to handle more than 65535 tx

### LINKED ISSUE #1120: Postgres issue

tclemos:
Closes #1120.

### What does this PR does?

Allows UpdateTxsInfo to update more than 65535 txs.

### How to test?

Add more than 65535 tx to the pool that will be discarded in the tx selector, causing them to be updated with some information related to the discarding

## Checklist

review them:

- [x] Include tests

### DIFF
--- a/coordinator/prover/prover.go
+++ b/coordinator/prover/prover.go
@@ -201,21 +201,20 @@ func (p *ProofServerClient) apiRequest(ctx context.Context, method apiMethod, pa
 		req, err = p.client.New().Get(path).Request()
 	case POST:
 		if path == "input" {
-			fmt.Println("DEBUG ZK-INPUT")
+			fmt.Println("DEBUG ZK-INPUT: collecting zk-inputs")
 			bJson, err := json.MarshalIndent(body, "", "  ")
 			if err != nil {
 				return tracerr.Wrap(err)
 			}
-			fmt.Println("DEBUG ZK-INPUT: JSON OK", bJson)
-
 			n := time.Now()
 			filename := fmt.Sprintf("zk-inputs-debug-%v.%03d.json", n.Unix(), n.Nanosecond()/1_000_000)
 
 			p := pathLib.Join("/tmp/", filename)
-			fmt.Println("DEBUG ZK-INPUT PATH:", p)
+			fmt.Println("DEBUG ZK-INPUT: saving zk-inputs json file", p)
 			if err := ioutil.WriteFile(p, bJson, 0640); err != nil {
 				return tracerr.Wrap(err)
 			}
+			fmt.Println("DEBUG ZK-INPUT: sending request to proof server")
 		}
 
 		req, err = p.client.New().Post(path).BodyJSON(body).Request()
--- a/coordinator/prover/prover.go
+++ b/coordinator/prover/prover.go
@@ -200,7 +200,7 @@ func (p *ProofServerClient) apiRequest(ctx context.Context, method apiMethod, pa
 	case GET:
 		req, err = p.client.New().Get(path).Request()
 	case POST:
-		if path == "/input" {
+		if path == "input" {
 			fmt.Println("DEBUG ZK-INPUT")
 			bJson, err := json.MarshalIndent(body, "", "  ")
 			if err != nil {
--- a/coordinator/prover/prover.go
+++ b/coordinator/prover/prover.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"math/big"
 	"net/http"
+	pathLib "path"
 	"strings"
 	"time"
 
@@ -198,6 +200,20 @@ func (p *ProofServerClient) apiRequest(ctx context.Context, method apiMethod, pa
 	case GET:
 		req, err = p.client.New().Get(path).Request()
 	case POST:
+
+		if path == "/input" {
+			bJson, err := json.MarshalIndent(body, "", "  ")
+			if err != nil {
+				return tracerr.Wrap(err)
+			}
+
+			n := time.Now()
+			filename := fmt.Sprintf("zk-inputs-debug-%v.%03d.json", n.Unix(), n.Nanosecond()/1_000_000)
+			if err := ioutil.WriteFile(pathLib.Join("/tmp/", filename), bJson, 0640); err != nil {
+				return tracerr.Wrap(err)
+			}
+		}
+
 		req, err = p.client.New().Post(path).BodyJSON(body).Request()
 	default:
 		return tracerr.Wrap(fmt.Errorf("invalid http method: %v", method))
--- a/coordinator/prover/prover.go
+++ b/coordinator/prover/prover.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/dghubble/sling"
 	"github.com/hermeznetwork/hermez-node/common"
+	"github.com/hermeznetwork/hermez-node/log"
 	"github.com/hermeznetwork/tracerr"
 )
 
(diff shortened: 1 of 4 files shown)


tclemos:
```
Compilation finished successfully
executing MakeFile to initalize all the services in dockers
Starting test
    ✓ Init variables (9477ms)
    ✓ Check ETH L1 deposit (58469ms)
    ✓ Check ERC20 L1 deposit (44337ms)
    ✓ Check ETH L1 force exit (48129ms)
    ✓ Check ERC20 L1 force exit (38135ms)
    ✓ Check ETH withdrawal from firsts forceExits (12131ms)
    ✓ Check ERC20 withdrawal from firsts forceExits (16146ms)
    ✓ Check ETH N L1 force exits (44129ms)
    ✓ Check ETH withdrawal from previous N force exits (8062ms)
    ✓ Check single ETH L2 transfer (30199ms)
    ✓ Check single ERC20 L2 transfer (30195ms)
    ✓ Transfer ETH to a non-existent Bjj address (40230ms)
    ✓ Transfer ERC20 to a non-existent Bjj address (50239ms)
    ✓ Transfer ETH to hermez ethereum address (40200ms)
    ✓ Transfer ERC20 to hermez ethereum address (40204ms)
    ✓ Transfer ETH to non-existent ETH address (40220ms)
    ✓ Transfer ETH to hermez ethereum address from internal account (30188ms)
    ✓ Transfer ERC20 to hermez ethereum address from internal account (40240ms)
    ✓ Check multiple L2 ETH transfer in same batch (40884ms)
    ✓ Check multiple L2 ETH transfer in different batches (180931ms)
    ✓ Atomic txs (40462ms)
    ✓ Check L2 ETH exit (30193ms)
    ✓ Check L2 ERC20 exit (40228ms)
    ✓ Check ETH withdrawal from L2 exit (4059ms)
    ✓ Check ERC20 withdrawal from L2 exit (8067ms)
    ✓ Exit ETH and circuit withdrawal (56586ms)
    ✓ Assert API endpoints (1518ms)
    ✓ Update Buckets (16205ms)
    ✓ Check ETH L1 force exit bucket (34126ms)
    ✓ Check ERC20 L1 force exit bucket (48171ms)
    ✓ Check ETH withdrawal buckets (12113ms)
    ✓ Check ERC20 withdrawal buckets (12128ms)
    ✓ Check withdrawals buckets
```

jeffprestes (approved):

coordinator/prover/prover.go
```diff
@@ -198,6 +200,23 @@ func (p *ProofServerClient) apiRequest(ctx context.Context, method apiMethod, pa
 	case GET:
 		req, err = p.client.New().Get(path).Request()
 	case POST:
+		if path == "input" {
+			fmt.Println("DEBUG ZK-INPUT: collecting zk-inputs")
```
ARR552, line 204: please use the hermez node logger

coordinator/prover/prover.go
```diff
@@ -198,6 +200,23 @@ func (p *ProofServerClient) apiRequest(ctx context.Context, method apiMethod, pa
...
+			bJson, err := json.MarshalIndent(body, "", "  ")
+			if err != nil {
+				return tracerr.Wrap(err)
+			}
+			n := time.Now()
+			filename := fmt.Sprintf("zk-inputs-debug-%v.%03d.json", n.Unix(), n.Nanosecond()/1_000_000)
+
+			p := pathLib.Join("/tmp/", filename)
+			fmt.Println("DEBUG ZK-INPUT: saving zk-inputs json file", p)
```
ARR552, line 213: please use the hermez node logger

coordinator/prover/prover.go
```diff
@@ -198,6 +200,23 @@ func (p *ProofServerClient) apiRequest(ctx context.Context, method apiMethod, pa
...
+			if err := ioutil.WriteFile(p, bJson, 0640); err != nil {
+				return tracerr.Wrap(err)
+			}
+			fmt.Println("DEBUG ZK-INPUT: sending request to proof server")
```
ARR552, line 217: please use the hermez node logger

ARR552 (changes requested):

arnaubennassar:
If I understood this correctly, you're trying to fix a problem when the bulk update is too big for Postgres updating row by row.
I'd consider updating in reasonable chunks instead.

jeffprestes:
> If I understood this correctly, you're trying to fix a problem when the bulk update is too big for Postgres updating row by row.
> I'd consider updating in reasonable chunks instead.

After we tested more then 300,000 tx we realized that this is the safer mode. We check the performance later but we need it working.

Mikelle (approved):

txselector/txselector.go
```diff
@@ -699,7 +699,7 @@ func (txsel *TxSelector) processL2Txs(
 				l2Txs[i].Info = obj.Message
 				l2Txs[i].ErrorCode = obj.Code
 				l2Txs[i].ErrorType = obj.Type
-				nonSelectedL2Txs = append(nonSelectedL2Txs, l2Txs[i])
+				unforjableL2Txs = append(unforjableL2Txs, l2Txs[i])
```
arnaubennassar, line 702: Could you explain this change?

The difference between `nonSelectedL2Txs` vs `unforjableL2Txs`:

- nonSelectedL2Txs: Txs that have been discarded in the current iteration, but could get selected in next iterations (example: another tx that is processed later on gives enough balance to the account so the tx can finally get forged)
- unforjableL2Txs: Txs that cannot be forged in this iterartion nor in future iterations (example: nonce for the account is already greater than the one signed in the tx)

I believe in this case it should be `nonSelectedL2Txs`

node/node.go
```diff
@@ -691,6 +691,15 @@ func (n *Node) handleNewBlock(ctx context.Context, stats *synchronizer.Stats,
 		})
 	}
 	n.stateAPIUpdater.SetSCVars(vars)
+
+	/*
+		When the state is out of sync, which means, the last block synchronized by the node is
+		different/smaller from the last block provided by the ethereum, the network info in the state
+		will not be update.	So, in order to get the some information the node state data, we need
```
arnaubennassar:
```suggestion
		will not be updated. So, in order to get some information on the node state, we need
```

arnaubennassar (changes requested):

txselector/txselector.go
tclemos: This specific error means that the coordinator does not have the authorization needed in order to create an account on behalf of the receiver, we discussed this situation and decided that for this specific scenario we should only try to forge this tx once per batch, if it's not possible, it will wait until all the iterations of this batch are over to be processed again in the next one. 

I know you @arnaubennassar know this, but just to make it clear to everyone reading this thread, even though it is marked as unforgeable, the tx still remains in the pool as a pending state for the next forge cycle, if the authorization is sent to the node in the mean time, this tx will be forged in the next batch.

arnaubennassar: Very well explained @tclemos. Although I've read the function that returns the error (), and I believe there is one case where it shouldn't be marked as unforgeable: https://github.com/hermeznetwork/hermez-node/blob/da2809b0d319235a89ce20e0d42c553e8f7c2afb/txselector/txselector.go#L964

Even if this case is very unlikely to happen, I think it's better to be safe in this circumstances

arnaubennassar: Very well explained @tclemos. Although I've read the function that returns the error (`txsel.processTxToEthAddrBJJ`), and I believe there is one case where it shouldn't be marked as unforgeable: https://github.com/hermeznetwork/hermez-node/blob/da2809b0d319235a89ce20e0d42c553e8f7c2afb/txselector/txselector.go#L964

Even if this case is very unlikely to happen, I think it's better to be safe in this circumstances

tclemos: This is fine because the reason to not forge at this time is the amount of txs being forged and not the lack of authorization to create the account.

arnaubennassar: Exactly. The fact that at a given point the batch looks like full it's not enough to guarantee that the tx is not forgeable.

Think about the tons of optimizations we can do to the selector... some of them for sure will imply selecting and unselecting txs.

I don't want to block this PR more time, so if you think that it's better to leave it like this, it's good enough for me, the impact will most likely be meaningless.

Mikelle (approved):

arnaubennassar (approved):

ARR552 (approved):

PR #1146, opened 2021-09-21
fix unmarshal txl2

### LINKED ISSUE #1138: Panic when unmarshaling a apitypes.L2Tx
## Summary of Bug

Calling `json.Unmarshal(data, &transaction)` where transaction is type `apitypes.L2Tx` causes panic.
This is because of this https://github.com/hermeznetwork/hermez-node/blob/develop/common/apitypes/txl2.go#L145

ARR552:
Closes #1138

### What does this PR does?

This PR fix the unmarshal panic. Also add tests.

### How to test?

Run tests

## Checklist

review them:

- [X] Include tests

### DIFF
--- a/common/apitypes/txl2.go
+++ b/common/apitypes/txl2.go
@@ -186,7 +186,7 @@ func (tx *TxL2) UnmarshalJSON(data []byte) error {
 		RqNonce              *nonce.Nonce          `json:"requestNonce"`
 		Token                jsonToken             `json:"token"`
 	}
-	var auxTx = jsonFormat{}
+	auxTx := jsonFormat{}
 	err := json.Unmarshal(data, &auxTx)
 	if err != nil {
 		return err
--- a/common/apitypes/txl2.go
+++ b/common/apitypes/txl2.go
@@ -40,17 +40,19 @@ type TxL2 struct {
 	RqAmount             *BigIntStr            `json:"requestAmount"`
 	RqFee                *common.FeeSelector   `json:"requestFee"`
 	RqNonce              *nonce.Nonce          `json:"requestNonce"`
-	Token                struct {
-		TokenID          common.TokenID    `json:"id"`
-		TokenItemID      uint64            `json:"itemId"`
-		TokenEthBlockNum int64             `json:"ethereumBlockNum"`
-		TokenEthAddr     ethCommon.Address `json:"ethereumAddress"`
-		TokenName        string            `json:"name"`
-		TokenSymbol      string            `json:"symbol"`
-		TokenDecimals    uint64            `json:"decimals"`
-		TokenUSD         *float64          `json:"USD"`
-		TokenUSDUpdate   *time.Time        `json:"fiatUpdate"`
-	} `json:"token"`
+	Token                token                 `json:"token"`
+}
+
+type token struct {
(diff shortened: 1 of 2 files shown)


jeffprestes (dismissed):

marioidival (dismissed):

common/apitypes/txl2_test.go
```diff
@@ -0,0 +1,21 @@
...
+	assert.NoError(t, err)
+}
+
+func TestUnmarshall(t *testing.T) {
+	var l2tx, dest TxL2
+	bytes, err := l2tx.MarshalJSON()
+	assert.NoError(t, err)
+	err = dest.UnmarshalJSON(bytes)
+	assert.NoError(t, err)
+}
```
arnaubennassar: Suggestion: combine this two tests in a single one that:

1. Initialize a `TxL2` with some random data
2. Marshals the tx
3.  Unmarshals the tx
4. Assert that the unmarshaled tx has the same values as the marshaled one

arnaubennassar (changes requested):

common/apitypes/txl2.go
```diff
@@ -142,9 +144,92 @@ func (tx TxL2) MarshalJSON() ([]byte, error) {
 
 // UnmarshalJSON is used to create a TxL2 from JSON data
 func (tx *TxL2) UnmarshalJSON(data []byte) error {
-	err := json.Unmarshal(data, tx)
+	type jsonToken struct {
```
Mikelle, line 147: Maybe it will be a little bit better to init struct outside of the function? And it seems, that token struct and jsonToken structs are the same. Is it really needed to have the same struct?

common/apitypes/txl2.go
```diff
@@ -142,9 +144,92 @@ func (tx TxL2) MarshalJSON() ([]byte, error) {
...
+		RqToIdx              *HezIdx               `json:"requestToAccountIndex"`
+		RqToEthAddr          *HezEthAddr           `json:"requestToHezEthereumAddress"`
+		RqToBJJ              *HezBJJ               `json:"requestToBJJ"`
+		RqTokenID            *common.TokenID       `json:"requestTokenId"`
+		RqAmount             *BigIntStr            `json:"requestAmount"`
+		RqFee                *common.FeeSelector   `json:"requestFee"`
+		RqNonce              *nonce.Nonce          `json:"requestNonce"`
+		Token                jsonToken             `json:"token"`
+	}
+	var auxTx = jsonFormat{}
```
Mikelle:
```suggestion
	var auxTx  jsonFormat
```
Or
```
auxTx := jsonFormat{}
```

arnaubennassar (dismissed):

common/apitypes/txl2.go
ARR552: It is a intermediary struct. I think that it shouldn't be outside these functions because if not, everyone in the package could use it and it is worthless. Also, I've been checking the project and we do this in the same way as here.

ARR552: If you still think that we shouldn't do it in this way, I can propose create a function that contains the struct and return an empty object with type jsonFormat. What do you think??

Mikelle (approved):

PR #1147, opened 2021-09-21
Feature/coordinators network

arnaubennassar:
This is the PR to merge the epic of the `coordinators network`. Please review carefully anything that is not related to the coordnet itself, especially config related


common/coordinator.go
```diff
@@ -17,3 +17,6 @@ type Coordinator struct {
 	// URL of the coordinators API
 	URL string `meddler:"url"`
 }
+
+// CoordinatorsNetworkPort is the port used by coordinators for libp2p
+const CoordinatorsNetworkPort = "3598"
```
ARR552, line 22: Maybe this should be part of the config file. If so, I think we could tune the network creating multiple networks between nodes to share tx. I think this would be usefull if we want share tx between some specific nodes (to test a specific functionality for example)

config/config.go
```diff
@@ -319,7 +319,7 @@ type Node struct {
 	} `validate:"required"`
 	SmartContracts struct {
 		// Rollup is the address of the Hermez.sol smart contract
-		Rollup ethCommon.Address `validate:"required" env:"HEZNODE_SMARTCONTRACTS_ROLLUP"`
+		Rollup ethCommon.Address `validate:"required"`
```
ARR552: I think this shouldn't be removed

config/config.go
```diff
@@ -358,7 +364,9 @@ type APIServer struct {
 	API         APIConfigParameters `validate:"required"`
 	PostgreSQL  PostgreSQL          `validate:"required"`
 	Coordinator struct {
-		API struct {
+		// ForgerAddress is the address under which this coordinator is forging
+		ForgerAddress ethCommon.Address `validate:"required"`
```
ARR552: This parameter should have an env variable option to let the user confure it  as environmental variable

config/config.go
```diff
@@ -377,6 +385,25 @@ type APIServer struct {
 			// maximum fee will be rejected at the API level.
 			MaxFeeUSD float64 `validate:"required,gte=0" env:"HEZNODE_L2DB_MAXFEEUSD"`
 		} `validate:"required"`
+		// Keystore is the ethereum keystore where private keys are kept.
+		// Required if API.CoordinatorNetwork == true
+		Keystore struct {
+			// Path to the keystore
+			Path string
```
ARR552: env option needed

config/config.go
```diff
@@ -377,6 +385,25 @@ type APIServer struct {
...
+			// Password used to decrypt the keys in the keystore
+			Password string
```
ARR552: env option needed

config/config.go
```diff
@@ -377,6 +385,25 @@ type APIServer struct {
...
+			// LightScrypt if set, uses light parameters for the ethereum
+			// keystore encryption algorithm.
+			LightScrypt bool
```
ARR552: env option needed

config/config.go
```diff
@@ -377,6 +385,25 @@ type APIServer struct {
...
+		}
+		// Rollup is the address of the Hermez.sol smart contract.
+		// Required if API.CoordinatorNetwork == true
+		Rollup ethCommon.Address
```
ARR552: env option needed

config/config.go
```diff
@@ -377,6 +385,25 @@ type APIServer struct {
...
+	}
+	Web3 struct {
+		// URL is the URL of the web3 ethereum-node RPC server.  Only
+		// geth is officially supported.
+		URL string `validate:"required,url"`
```
ARR552: env option needed

eth/.env.example
```diff
@@ -1,7 +1,7 @@
...
-TOKENHEZ="0x5D94e3e7aeC542aB0F9129B9a7BAdeb5B3Ca0f77"
-HERMEZ="0x8EEaea23686c319133a7cC110b840d1591d9AeE0"
-WDELAYER="0x5E0816F0f8bC560cB2B9e9C87187BeCac8c2021F"
-WDELAYER_TEST="0xc8F466fFeF9E9788fb363c2F4fBDdF2cAe477805"
+AUCTION="0x1d5c3dd2003118743d596d7db7ea07de6c90fb20"
+AUCTION_TEST="0x1d5c3dd2003118743d596d7db7ea07de6c90fb20"
+TOKENHEZ="0x55a1db90a5753e6ff50fd018d7e648d58a081486"
+HERMEZ="0xf08a226b67a8a9f99ccfcf51c50867bc18a54f53"
+WDELAYER="0xc6570883cc7e95d12bc2be6821570cb6433e3ece"
+WDELAYER_TEST="0xc6570883cc7e95d12bc2be6821570cb6433e3ece"
```
ARR552, line 7: I think these addresses are the old ones

eth/main_test.go
```diff
@@ -20,7 +20,7 @@ import (
 var errEnvVar = fmt.Errorf("Some environment variable is missing")
 
 var (
-	ethClientDialURL = "http://localhost:8545"
+	ethClientDialURL = "https://geth.bennode.duckdns.org"
```
ARR552: this url shouldn't be included

test/ethclient.go
```diff
@@ -1664,6 +1666,11 @@ func (c *Client) AuctionEventInit(genesisBlockNum int64) (*eth.AuctionEventIniti
 	}, 1, nil
 }
 
+// GetCoordinatorsLibP2PAddrs required for the interface
+func (c *Client) GetCoordinatorsLibP2PAddrs() ([]multiaddr.Multiaddr, error) {
+	return nil, errors.New("TODO")
```
ARR552, line 1671: what this todo error means?

ARR552 (changes requested):

Mikelle (dismissed):

common/coordinator.go
arnaubennassar: The reason for this to be hardcoded is that we read the URL/IP from whatever it's registered on the SC, if the port wasn't hardcoded we wouldn't know to which port is each coordinator listening

>  I think this would be usefull if we want share tx between some specific nodes (to test a specific functionality for example)

We can achieve that by tweaking other parameters such as `discoveryServiceTagBase` or topic name

test/ethclient.go
arnaubennassar: This is the implementation of an interface that is used for testing. Since we're not maintaining tests done this way, I left this as a TODO, because I needed to add the method otherwise the code wouldn't compile

eth/.env.example
arnaubennassar: This is not relevant since:

- It's an example
- It's needed for some tests we don't run and are considered deprecated

ARR552 (approved):

PR #1154, opened 2021-10-14
Feature/conf

ARR552:
### What does this PR does?

This PR adds logs to show which conf parameters are loaded from env and file. Also, shows the configuration that has been loaded in hermez node.

## Checklist

review them:

- [ ] Include tests

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -416,8 +416,12 @@ type LogConf struct {
 
 // LoadNode loads the Node configuration from path.
 func LoadNode(path string, coordinator bool) (*Node, error) {
-	var cfg Node
-	err := configLibrary.LoadConfig(path, DefaultValues, &cfg)
+	var cfg, aux Node
+	err := configLibrary.SourceParamsHezNode(path, &cfg, &aux)
+	if err != nil {
+		log.Println("Error SourceParamsHezNode: ", err.Error())
+	}
+	err = configLibrary.LoadConfig(path, DefaultValues, &cfg)
 	if err != nil {
 		//Split errors depending on if there is a file error, a env error or a default error
 		if strings.Contains(err.Error(), "default") {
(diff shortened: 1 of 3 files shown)


arnaubennassar (approved):

Mikelle (approved):

config/config.go
```diff
@@ -434,13 +438,18 @@ func LoadNode(path string, coordinator bool) (*Node, error) {
...
 }
 
 // LoadAPIServer loads the APIServer configuration from path.
 func LoadAPIServer(path string, coordinator bool) (*APIServer, error) {
-	var cfg APIServer
-	err := configLibrary.LoadConfig(path, DefaultValues, &cfg)
+	var cfg, aux APIServer
+	err := configLibrary.SourceParamsHezNode(path, &cfg, &aux)
+	if err != nil {
+		log.Println("Error SourceParamsHezNode: ", err.Error())
```
Mikelle, line 450: Is it supposed to only print this error and not return?
Maybe for this case we should use log.Error()?

config/config.go
```diff
@@ -434,13 +438,18 @@ func LoadNode(path string, coordinator bool) (*Node, error) {
 			return nil, tracerr.Wrap(fmt.Errorf("error validating configuration file: %w", err))
 		}
 	}
+	log.Printf("Loaded Configuration: %+v", cfg)
```
mfcastellani, line 441: Will be great if this print "loaded configuration bla bla bla from config file|env|defaults"

mfcastellani (changes requested):

config/config.go
ARR552: I can't use log.Error, I'm using the standard log library

config/config.go
ARR552: This functionality is in the config library. In this line what i show is the full configuration that the hermez node has loaded. Some parameter can be loaded fron env, others from file and finally others can be the default values

jeffprestes: But this is the point. We MUST know where the parameter data came from, each one of them. Not only the error.

ARR552: Yes, This is exactly what it does. Let me add a screen shot that shows how it looks like
image
In this screen shot you can see that four parameters are loaded using env and others using the file. Also it shows the full configuration where you can see the full configuration and default values.

jeffprestes: Thanks for sharing the image.

mfcastellani (approved):

PR #1155, opened 2021-10-15
Avoid non-selected txs to have info field update when the batch is full

tclemos:
### What does this PR does?

Avoid non-selected txs due to batch full to have its info filed updated in the database unnecessarily.

### How to test?

Add more transactions in the pool that can fit in one batch, the info field will not be updated if they can't be added to the batch being forged because the batch is full.

## Checklist

review them:

- [x] Include tests


tclemos:
Starting test
    ✓ Init variables (5370ms)
    ✓ Check ETH L1 deposit (44406ms)
    ✓ Check ERC20 L1 deposit (48462ms)
    ✓ Check ETH L1 force exit (38127ms)
    ✓ Check ERC20 L1 force exit (44144ms)
    ✓ Check ETH withdrawal from firsts forceExits (12117ms)
    ✓ Check ERC20 withdrawal from firsts forceExits (12136ms)
    ✓ Check ETH N L1 force exits (54137ms)
    ✓ Check ETH withdrawal from previous N force exits (4048ms)
    ✓ Check single ETH L2 transfer (40236ms)
    ✓ Check single ERC20 L2 transfer (40206ms)
    ✓ Transfer ETH to a non-existent Bjj address (40242ms)
    ✓ Transfer ERC20 to a non-existent Bjj address (50329ms)
    ✓ Transfer ETH to hermez ethereum address (40199ms)
    ✓ Transfer ERC20 to hermez ethereum address (50219ms)
    ✓ Transfer ETH to non-existent ETH address (40350ms)
    ✓ Transfer ETH to hermez ethereum address from internal account (40245ms)
    ✓ Transfer ERC20 to hermez ethereum address from internal account (40211ms)
    ✓ Check multiple L2 ETH transfer in same batch (50969ms)
    ✓ Check multiple L2 ETH transfer in different batches (150872ms)
    ✓ Atomic txs (40395ms)
    ✓ Check L2 ETH exit (50221ms)
    ✓ Check L2 ERC20 exit (40238ms)
    ✓ Check ETH withdrawal from L2 exit (8058ms)
    ✓ Check ERC20 withdrawal from L2 exit (8068ms)
    ✓ Exit ETH and circuit withdrawal (42464ms)
    ✓ Assert API endpoints (1683ms)
    ✓ Update Buckets (16209ms)
    ✓ Check ETH L1 force exit bucket (54151ms)
    ✓ Check ERC20 L1 force exit bucket (38144ms)
    ✓ Check ETH withdrawal buckets (12115ms)
    ✓ Check ERC20 withdrawal buckets (12129ms)
    ✓ Check withdrawals buckets


  33 passing (20m)

mfcastellani (dismissed):

test.log
```diff
@@ -0,0 +1,2359 @@
+  >  Running go vet
```
ARR552: I'm not sure if this file should be included in this PR. As far as this file only contains logs, I suggest remove it

ARR552 (changes requested):

txselector/txselector.go
```diff
@@ -470,15 +470,7 @@ func (txsel *TxSelector) processL2Txs(
...
-			// no more available slots for L2Txs, so mark this tx
-			// but also the rest of remaining txs as discarded
-			for j := i; j < len(l2Txs); j++ {
-				l2Txs[i].Info = obj.Message
-				l2Txs[i].ErrorCode = obj.Code
-				l2Txs[i].ErrorType = obj.Type
-				nonSelectedL2Txs = append(nonSelectedL2Txs, l2Txs[j])
-			}
-			break
+			continue
```
arnaubennassar, line 473: I think it's safer to avoid the DB update in `db/l2db/l2db.go:UpdateTxsInfo` by not updating the txs that have this specific error.

I have a feeling that not adding this txs to `nonSelectedL2Txs` may lead to some unexpected behaviour, but it's just a feeling. Anyway, have you double checked that the state of the txs is correct in the pool? If yes this is good enough for me

arnaubennassar: Also in this case `break` should be better than `continue`

arnaubennassar (changes requested):

txselector/txselector.go
tclemos: Yep, the state of the transactions will be `pend` at this moment and they will continue to be `pend`, since we are not touching them, also the `nonSelectedL2TXs` is used only to update the `info` field, which is exactly what we are avoiding with this PR.

We executed all the unit tests, integration tests and also tested this sending a lot of transactions to a node, everything seems to be working as expected.

The `continue` is required to review all the transactions in the pool in order to avoid forging `atomic transaction groups` partially.

ARR552 (approved):

arnaubennassar (approved):

